PR #698, opened 2017-02-24
Iter

kaeluka:
This is starting to be useful, that's why I'll increase visibility by sending a PR.

There are examples in the tests that show how for comprehensions can be used.

Contributions:

 - For comprehensions with support for filtering (`when` clauses)
 - Native arrays, Maybe types, Futures, now implement an extensible set of methods without boxing. Check out the documentation of the `modules/BuiltinExt` module.
 - An iterator library (design strongly following Rust's lib) that can be used for filtering.
 - Native arrays support internal (via map/flatMap/filter) iteration and external iteration (via iterators).


TobiasWrigstad (changes requested):
This is a high-level review with lots of discussion points. I had hoped to respond with a PR on the PR fixing some of these things, but I'm currently too busy with other things :(

In general I think we should retire the old `for` and unify `For` and `Foreach` and rename them `for` before we merge this. 

Also, if it is possible to get the inference working, that would be nice so we don't have to make many changes in the code that uses for loops. 

We can use stuff from the #701 to check whether a return value is needed or not and use that to speed up comprehensions that do not need a result. 

We should also provide some standard boxes in the standard library. Possibly in Builtin.

PR #746, opened 2017-03-23
Extend forward

PhucVH888:
### Introduction
```
2. forward FutureChaining
3. forward PreexistingFuture
4. TODO: forward in closure's body
```
The first case has been solved in PR #682, and the last case will be done soon.
In this PR, the `forward` can deal with future chaining (case 2), and preexisting future (case 3).
### Syntax
```
2. forward((new Foo)!foo() ~~> \(x: t) -> …)
3. forward(f)
```
#### 2. `forward FutureChaining`
```
      forward (x ! foo() ~~> \lam)
```
where `f` is the current future
```
| calling | result | semantics                              |
| async   | yes    | f' = new Fut; x ! foo(f'); \           |
|         |        |   f'.chain(f, \lam); return            |
| async   | no     | f' = new Fut; x ! foo(f'); \           |
|         |        |   f',chain(null, \lam); return         |
| sync    | yes    | f' = new Fut; x ! foo(f'); \           |
|         |        |   f'' = new Fut(); f'.chain(f'',\lam); \ |
|         |        | return get f''                         |
| sync    | not    | f' = new Fut; x ! foo(f'); \           |
|         |        |   f'.chain(null,\lam); return          |
```
#### 3. `forward PreexistingFuture`
```
      forward (f)
```
where `f` is the preexisting future. The `foward(f)` will be desugared as `foward(f ~~> id)`
```
class Base
  def one(arg : Fut[int]) : int
    forward(arg)
  end
end
```
Another example is to chain on preexisting future.
```
class Base
  def one(arg : Fut[int]) : int
    forward(arg ~~> fun(x:int) => x * 10)
  end
end
```
### Examples


supercooldave:
This has been neglected for too long.
Please rebase. If it has any more functionality than previous version, make this PR deal with that. 
If the PR looks incomplete (e.g., it includes TODOs), then it won't receive a proper review.

PhucVH888:
Temporarily close.

PR #756, opened 2017-03-28
Sophia's Barbers work

sophiaIC:
Make sure these boxes are checked before submitting your pull request:

-  This is a SleepingBarber implementation
- I ran several tests, no crash, and output looked reasonable
-  corresponds to saving Akka version


supercooldave:
I tried to fix this, both using the web editor and by checking out the PR, but I cannot. It seems that the reason is @sophiaIC encore repo is private. 

@sophiaIC you need to fix the conflict (which is in the README file) and repush, otherwise I fear that this might be lost.

supercooldave:
Closing as PR is broken (since going public) and out of date (not Kappa compliant).
New PR replacing this will follow soon.

PR #767, opened 2017-03-31
added 13 and 14

sophiaIC:
- parts 13 and 14 of the savina suite
- tested with various inputs
- they compile and do not crash


supercooldave:
Closing as PR is broken (since going public) and out of date (not Kappa compliant).
New PR replacing this will follow soon.

Issue #770, opened 2017-04-06
Linear object in Tuple can't be consumed


kaeluka:
I have a function `foo() : (something, somethingElse)`, where `somethingElse` is a `linear class`; I’d like to do this:

```
match foo() with
  case (a, b) =>
    ... consume b ...
```

I can’t do this, because the `b` binding is immutable, so the `consume` will not typecheck.

supercooldave:
I expect that `(something, somethingElse)` should *not* be well-formed unless it is `linear`.

supercooldave:
Arrrghhhhh. I have this problem, but with triples.

PR #771, opened 2017-04-06
Promise lib

kaeluka:
This contributes a promise lib. It will be useful to the Async IO project.


kikofernandez:
There is already a [`Promise` class in the standard library](https://github.com/parapluu/encore/blob/development/modules/standard/Task.enc). I think both of them are quite similar, except that one uses an active class and this one uses the `unsafe`.

modules/standard/Data/Promise.enc
```diff
@@ -0,0 +1,32 @@
...
+-- Once you call fulfil(..) on the promise object, the
+-- future returned by this call will be fulfilled.
+fun mkPromise[t]() : (Fut[t], Promise[t])
+  val f = EMBED (Fut[t]) future_mk(_ctx, _enc__type_t); END
+  (f, new Promise(f))
+end
+
+-- This should be linear, but we're hitting issue #770 then
+-- in the mean time, do not alias promises!
+unsafe class Promise[t]
```
supercooldave, line 15: I wonder whether this can be refactored into two traits, one for the Fut and one linear one for handling Fulfilling. Then you could have `class Promise : Fut + linear Fulfiller` (or whatever) and can use `new Promise` to create instances, rather than the somewhat strange `mkPromise`.

supercooldave (commented):
Comment about general structure of proposed construct.

modules/standard/Data/Promise.enc
TobiasWrigstad: That's what I did for my userland implementation of futures that no one liked

supercooldave: @TobiasWrigstad What exactly do you mean `that`? Is that referring to the trait-based decomposition. There was no linearity then.
Also, I don't think my proposal is good enough. You want to consume the promise after fulfilling that. Probably needs a function.

kaeluka: I had no idea that we already had promises :-(

Why are they implemented using suspend? @supercooldave's idea about user trait decomposition is interesting, but give that we have promises already it seems unlikely that this PR will be merged.

supercooldave: I suspect there is a superior implementation of promises waiting to be implemented that can exploit the features of Encore and doesn't use `suspend`. 
If this current version is implemented to support a student's project, then it can live in the student's repo for now.

kaeluka: > If this current version is implemented to support a student's project, then it can live in the student's repo for now.

Yes, I think so, too!
ping @DumKatt

TobiasWrigstad:
@kikofernandez Can we now safely rewrite the promise using forward instead of suspend?

modules/standard/Data/Promise.enc
TobiasWrigstad: > I suspect there is a superior implementation of promises waiting to be implemented that can exploit the features of Encore and doesn't use suspend.

For sure. I believe we can change line 108 of the standard promises to use forward instead of suspend. Much more efficeint. 

```
def deliver() : a
  if this.fulfilled() do
    match this.value with
      case Just(v) => v
      case _       => abort("This should not happen!")
    end
  else
    forward this ! deliver()
  end
end
```

kikofernandez:
@TobiasWrigstad I think we can but I am not sure if we need things from #746. As far as I know, we don't have many (if any) user of the promise library.

TobiasWrigstad:
@kikofernandez -- please see the code I pasted in thread above. Should make do with what's in #682. 

Einar and I are writing an object groups implementation that is using the promises, but that's vapor ware. 
I also think you are using it in the par t code you were writing.

modules/standard/Data/Promise.enc
supercooldave: Better still would be to do `forward (f ~~> id)` where `f` is the underlying future and `id` is the identity function. This would avoid the spin lock (finger crossed).

TobiasWrigstad:
Of course! But that stuff is not yet merged right?


7 apr. 2017 kl. 18:50 skrev supercooldave <notifications@github.com<mailto:notifications@github.com>>:


@supercooldave commented on this pull request.

________________________________

In modules/standard/Data/Promise.enc<https://github.com/parapluu/encore/pull/771#discussion_r110433720>:

> @@ -0,0 +1,32 @@
+-- Promises are futures that you can fulfil by explictly calling a method that you pass a value.
+module Promise
+
+-- Create a promise, together with a future that can be
+-- fulfilled using the promise:
+-- Once you call fulfil(..) on the promise object, the
+-- future returned by this call will be fulfilled.
+fun mkPromise[t]() : (Fut[t], Promise[t])
+  val f = EMBED (Fut[t]) future_mk(_ctx, _enc__type_t); END
+  (f, new Promise(f))
+end
+
+-- This should be linear, but we're hitting issue #770 then
+-- in the mean time, do not alias promises!
+unsafe class Promise[t]


Better still would be to do forward (f ~~> id) where f is the underlying future and id is the identity function. This would avoid the spin lock (finger crossed).

supercooldave:
Closing until a better version of the existing Promise lib is available.

PR #772, opened 2017-04-07
Support for borrow blocks

EliasC:
This PR adds support for borrow blocks:

```
borrow e as x in -- e : t
  ... -- x : borrowed t
end
```

When `e` is linear, the whole path must consist only of linear
targets (same rule as for `borrowed` function parameters), and the root
of the path may not be accessed inside the block.

The desugaring of `?.` and `?!` has been changed to first borrow the
target, so that it may be used on linear targets as well. This was the
main motivation for this addition.


supercooldave:
Accepting the bounded mode polymorphism PR broke this.
If you fix that bug, I can review this.

supercooldave:
I'm happy with this. The code is nicely written and it is extensively tested.

PR #773, opened 2017-04-07
Remove unused ParT-nodes

EliasC:
These nodes were only used for `OldParser.hs`, which has since been
removed. They are therefore dead code.


kikofernandez (approved):
This refactoring looks good and it was about time :)
Merging today around 13.00

PR #774, opened 2017-04-07
Add savina 10th concurrent sorted list

albertnetymk:


supercooldave:
Merging this evening if there are no objections.

src/tests/stress/savina/10.ConcurrentSortedLinkedList/Main.enc
```diff
@@ -0,0 +1,171 @@
...
+    this.i = i
+  end
+
+  def compare(other:Ord) : int
+    val y = EMBED (Integer) (void*)#{other}; END
+    this.i - y.i
+  end
+end
+
+local class Node : Id
```
EliasC: Could this class be made `subord` rather than `local`? It would make sure that the nodes are kept local to each list.

src/tests/stress/savina/10.ConcurrentSortedLinkedList/Main.enc
```diff
@@ -0,0 +1,171 @@
...
+
+read class Integer : Ord(i)
+  val i:int
+
+  def init(i:int) : unit
+    this.i = i
+  end
+
+  def compare(other:Ord) : int
+    val y = EMBED (Integer) (void*)#{other}; END
```
EliasC, line 20: You could get rid of this `EMBED` if you wrote `Ord` using the following pattern:
```
read trait Ord[self]
  require def compare(x : self) : int
end

read class Integer : Ord[Integer](i)
  val i:int

  def init(i:int) : unit
    this.i = i
  end

  def compare(other:Integer) : int
    this.i - other.i
  end
end
```

albertnetymk:
Only managed to apply the first change.

What should `val elem : Ord` be updated after the second change? @EliasC

EliasC:
You would have to change `class Node` to `class Node[t]` and make `val elem : Ord[t]`. Similarly you probably have to change the list to `SortedList[t]`, and instantiate it as `SortedList[Integer]`. Maybe this means that you need to change where the `Integer`s are created. If this doesn't work it's fine to leave it like this for now.

When we have bounded polymorphism (in the pipeline) you would write `class Node[t : Ord]` and have `val elem : t`.

EliasC:
@supercooldave I'm fine with merging this as soon as Albert gives a 👍 , regardless of the change to `Node` (as a better change will be possible soon).

albertnetymk:
The workaround due to lack of bounded polymorphism seems quite invasive. I prefer the current implementation, where `EMBED` is localized.

supercooldave:
@albertnetymk So I shouldn't expect any changes, and merge as is?

albertnetymk:
No more changes from me unless there are further comments.

PR #776, opened 2017-04-07
Bounded mode polymorphism

EliasC:
This commit adds support for adding bounds on the mode of a type
parameter. It also introduces the mode `sharable` which abstracts
over all modes that may be safely shared between active
objects (`read`, `active` and `shared`). This mode can only be
used for type parameters for now, where it allows a polymorphic
context to pass values between active objects. See the changed/added
tests for more details.


src/types/Typechecker/TypeError.hs
```diff
@@ -763,6 +775,16 @@ instance Show Error where
         printf ("Method '%s' returns a local capability and cannot " ++
                 "be called by a different active object")
                (show name)
+    show (PolymorphicArgumentSendError arg ty) =
+        printf ("Cannot pass polymorphic expression '%s' between active " ++
```
supercooldave: "expression" might not be the right word. You cannot pass an expression.

src/types/Typechecker/TypeError.hs
```diff
@@ -799,9 +821,17 @@ instance Show Error where
     show (ActiveTraitError trait) =
         printf "Trait '%s' can only be included by active classes"
                (show trait)
-    show (UnsafeTypeArgumentError ty) =
-        printf "Cannot use non-aliasable type '%s' as type argument."
-               (show ty)
+    show (UnsafeTypeArgumentError formal ty) =
+        if isModeless ty then
+          printf "Cannot use non-aliasable type '%s' as type argument" --TODO: This seems imprecise...
```
supercooldave: Is non-aliasable the same as linear?

supercooldave (commented):
Minor comments for now.

modules/standard/Active/ActiveTest.enc
```diff
@@ -1,12 +1,13 @@
--- This file was automatically converted by encorec
-
 import Active.Active
```
supercooldave, line 1: Shouldn't this test be moved to the test directory?
(Yes, you can argue that you didn't put it here, but as you are modifying the test, then now is the time!!)

supercooldave:
I'm pretty happy with this. My feeling is that we should merge it, after the very minor comments are addressed.
Does anyone else want to look at it?

src/types/Typechecker/TypeError.hs
EliasC: "Cannot pass polymorphic value"?

src/types/Typechecker/TypeError.hs
EliasC: The `TODO` there is relevant... I'll look into when this case is hit exactly and change it accordingly. It might be that this is actually dead code.

modules/standard/Active/ActiveTest.enc
EliasC: I would even argue that it *is* my responsibility, as it's PR that makes this library (verifiably) safe to use again.

src/types/Typechecker/TypeError.hs
supercooldave: Or more verbosely, "values of polymorphic type".

src/types/Typechecker/TypeError.hs
supercooldave: Let me know.

modules/standard/Active/ActiveTest.enc
supercooldave: Let me know.

src/types/Typechecker/TypeError.hs
EliasC: > Cannot pass value of '%s' between active objects. Its type is polymorphic so it may not be safe to share.

supercooldave: Looks good.

src/types/Typechecker/TypeError.hs
EliasC: Actually, the code here makes sense, albeit it's not very precise. This branch is for when we have a type that is indirectly `linear`, for example a tuple where one element is `linear`. It doesn't say "linear" in the error message as it gives the same error for `subord`inate types. I've kept the TODO, but elaborated on it a bit to make it more clear how the error message might be improved in the future.

modules/standard/Active/ActiveTest.enc
EliasC: Moved!

EliasC:
I've reacted to Dave's comments.

modules/standard/Active/ActiveTest.enc
supercooldave: Has it moved?

EliasC: Hm... Copied seems more correct. Removing the old one now!

EliasC:
Pushed additional changes:
* Changed the parser so that it only allows modes for type variables when they appear in type parameter lists
* Actually moved `ActiveTest.enc`

PR #777, opened 2017-04-11
Add savina benchmark 4. Fork Join Throughput

kikofernandez:
Add savina benchmark `4. Fork Join Throughput`

### DIFF
--- a/src/tests/stress/savina/4.fjthrput/ThroughputActor.enc
+++ b/src/tests/stress/savina/4.fjthrput/ThroughputActor.enc
@@ -1,0 +1,18 @@
+module ThroughputActor
+
+import ThroughputConfig
+
+active class ThroughputActor
+  var messagesProcessed: uint
+  val totalMessages : uint
+
+  def init(totalMessages: int): unit
+    this.messagesProcessed = 0
+    this.totalMessages = totalMessages
+  end
+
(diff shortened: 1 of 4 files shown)


kikofernandez:
I have assigned the reviewer as per [README.md in Savina](https://github.com/parapluu/encore/tree/development/src/tests/stress/savina)

supercooldave:
Again, I don't see any joining.

kikofernandez:
again, it's a direct translation. You should check the original code and the one produced here.
(I am not making this up, it's the given name and a direct translation)

supercooldave:
`AkkaActorState.awaitTermination(system)`

kikofernandez:
reading the program carefully, I understand that they send messages and need to wait until all actors have finished their work. This is implicit in Encore, as the program cannot terminate unless all actors have performed their work (copy-paste from #778 )

supercooldave (changes requested):
This can be accepted, IMO, after changing the name of the directory 4.fjthrput to something more parseable, such as 4.ForkJoinThroughput.

kikofernandez:
the reason the naming was to have the same name convention as the original code, so that it's easier to see which benchmarks we have implemented. I am happy to do the renaming.

supercooldave:
Best not change the name then.

PR #778, opened 2017-04-11
Added the forkjoin benchmark

kikofernandez:
Added the forkjoin benchmark

### DIFF
--- a/src/tests/stress/savina/5.fjcreate/ForkJoinConfig.enc
+++ b/src/tests/stress/savina/5.fjcreate/ForkJoinConfig.enc
@@ -1,0 +1,21 @@
+module ForkJoinConfig
+
+read class ForkJoinConfig
+  val N: int
+  val C: int
+
+  def init(N: int, C: int): unit
+    this.N = N
+    this.C = C
+  end
+
+  def performComputation(theta: real): unit
+    val sint = EMBED (real)
+                 sin(#{theta});
+               END
+    val res = sint * sint
+    if (res <= 0) then
+      abort("IllegalStateException: Benchmark exited with unrealistic res value")
+    end
+  end
+end
--- a/src/tests/stress/savina/5.fjcreate/Main.enc
+++ b/src/tests/stress/savina/5.fjcreate/Main.enc
@@ -8,10 +8,6 @@ fun getFromMaybe(x: Maybe[int]): int
   end
 end
 
-fun bool_from_int(x: int): bool
-  not (x == 0)
-end
-
 active class Main
   def parseArgs(args: [String]): ForkJoinConfig
     if |args| > 1 then
--- a/src/tests/stress/savina/5.fjcreate/Main.enc
+++ b/src/tests/stress/savina/5.fjcreate/Main.enc
@@ -9,11 +9,7 @@ fun getFromMaybe(x: Maybe[int]): int
 end
 
 fun bool_from_int(x: int): bool
-  if x == 0 then
-    return false
-  else
-    return true
-  end
+  not (x == 0)
 end
 
 active class Main
--- a/src/tests/stress/savina/5.fjcreate/Main.enc
+++ b/src/tests/stress/savina/5.fjcreate/Main.enc
@@ -1,0 +1,41 @@
+import ForkJoinConfig
+import ForkJoinActor
+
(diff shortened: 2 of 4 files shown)


kikofernandez:
I have assigned the reviewer as per [README.md in Savina](https://github.com/parapluu/encore/tree/development/src/tests/stress/savina)

supercooldave:
I don't see any join in this framework, just forking.

kikofernandez:
true, but that's the Savina name and the direct [translation of this benchmark](https://github.com/shamsimam/savina/blob/master/src/main/scala/edu/rice/habanero/benchmarks/fjcreate/ForkJoinAkkaActorBenchmark.scala)

supercooldave:
This line
```
  AkkaActorState.awaitTermination(system)
```
seems to be `join`.

kikofernandez:
reading the program carefully, I understand that they send messages and need to wait until all actors have finished their work. This is implicit in Encore, as the program cannot terminate unless all actors have performed their work

supercooldave:
I won't fight this.
Were the original benchmarks done for Akka? If not, I wonder what they look like.

kikofernandez:
I didn't get the comment at first and we have discussed this offline. 
The question refers to whether other libraries implement this in the same way. 
If we look at the [benchmark using Habanero](https://github.com/shamsimam/savina/blob/master/src/main/scala/edu/rice/habanero/benchmarks/fjcreate/ForkJoinHabaneroActorBenchmark.scala), they do exactly the same thing I did (no `join` operation involved)

src/tests/stress/savina/5.fjcreate/ForkJoinConfig.enc
```diff
@@ -0,0 +1,21 @@
...
+    this.N = N
+    this.C = C
+  end
+
+  def performComputation(theta: real): unit
+    val sint = EMBED (real)
+                 sin(#{theta});
+               END
+    val res = sint * sint
+    if (res <= 0) then
```
supercooldave, line 17: sin(0) * sin(0) = 0
Why is this an unrealistic value?

src/tests/stress/savina/5.fjcreate/Main.enc
```diff
@@ -0,0 +1,41 @@
...
+import ForkJoinActor
+
+fun getFromMaybe(x: Maybe[int]): int
+  match x with
+    case Just(i) => i
+    case Nothing => abort("Error: no value in maybe type")
+  end
+end
+
+fun bool_from_int(x: int): bool
```
supercooldave: why not just `not x == 0`

supercooldave (changes requested):
This will be ready to go after the minor comments have been addressed, and after changing benchmark name from `5.fjcreate` to `5.ForkJoinCreate` (or something appropriate).

src/tests/stress/savina/5.fjcreate/Main.enc
kikofernandez: > operator 'not' is not defined for values of type 'int'

src/tests/stress/savina/5.fjcreate/ForkJoinConfig.enc
kikofernandez: I didn't make this up :)
[Original code here](https://github.com/shamsimam/savina/blob/master/src/main/java/edu/rice/habanero/benchmarks/fjcreate/ForkJoinConfig.java#L47)

```scala
protected static void performComputation(final double theta) {
    final double sint = Math.sin(theta);
    final double res = sint * sint;
    //defeat dead code elimination
    if (res <= 0) {
        throw new IllegalStateException("Benchmark exited with unrealistic res value " + res);
    }
}
```

src/tests/stress/savina/5.fjcreate/Main.enc
supercooldave: Must be `not (x == 0)`

src/tests/stress/savina/5.fjcreate/ForkJoinConfig.enc
supercooldave: Amazing! Someone doesn't know his math.
Okay, just leave it.

kikofernandez:
i have removed the function `bool_from_int, it was not used

PR #766, opened 2017-03-31
Tidied up Chameneos

supercooldave:
While checking that Chameneos benchmark was okay, I tidied the code up a little.

PR #779, opened 2017-04-11
Chameneos

supercooldave:
While checking that Chameneos benchmark was okay, I tidied the code up a little. (Now Updated).


EliasC:
I already reviewed this PR in in #766. Merging as soon as CI has finished!

PR #780, opened 2017-04-11
Allow active traits in modeless classes

EliasC:
The semantics still only allows active traits to be composed with
other active traits, but with this fix you don't have to make the
class itself `active`. The test for active traits has been adapted.

In other words, the following is now allowed:
```
trait T
  ...
end

class C : active T
  ...
end
```

### DIFF
--- a/src/back/CodeGen/Header.hs
+++ b/src/back/CodeGen/Header.hs
@@ -263,8 +263,8 @@ generateHeader p =
                  mparams = A.methodParams m
                  mtype = A.methodType m
 
-     wrapperMethods A.Class{A.cname, A.cmethods} =
-       if Ty.isPassiveRefType cname then
+     wrapperMethods c@A.Class{A.cname, A.cmethods} =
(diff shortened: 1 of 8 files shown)


supercooldave:
Bleh!

EliasC:
@supercooldave: Why "Bleh"?

EliasC:
You can do this with all the other modes, so this only makes things more regular.

supercooldave:
I like being up front about activeness.

> active class Foo

or 

> actor Foo

rather than being less so. 

I guess I don't have to use it.

EliasC:
> I like being up front about activeness.

Me too! But when writing the documentation for traits and capabilites, it seemed unintuitive to have a special case for active traits (i.e. only allowing them when the class is active). As soon as I get to documenting the classes, I'm going to suggest using modes whenever you want a class that does nothing "fancy".

supercooldave:
Looks good.

PR #783, opened 2017-04-12
Renamed field to remove warning

supercooldave:
This PR renames a field in one of the class `Synchroniser` in the `Task` module to eliminate the following warning:

> Warning at "~/encore/modules/standard/Task.enc" (line 129, column 3):
> Field 'result' holds an array and could be confused with the method of the same name


EliasC:
Looks straightforward enough! Merging later today.

PR #784, opened 2017-04-13
Extended forward

PhucVH888:
Re-created the PR from orphan PR #746 after deleting my repo and forked a new one.

### Introduction
```
2. forward FutureChaining
3. forward PreexistingFuture
4. TODO: forward in closure's body
```
The first case has been solved in PR #682, and the last case will be done soon.
In this PR, the `forward` can deal with future chaining (case 2), and preexisting future (case 3).
### Syntax
```
2. forward((new Foo)!foo() ~~> \(x: t) -> …)
3. forward(f)
```
#### 2. `forward FutureChaining`
```
      forward (x ! foo() ~~> \lam)
```
where `f` is the current future
```
| calling | result | semantics                              |
| async   | yes    | f' = new Fut; x ! foo(f'); \           |
|         |        |   f'.chain(f, \lam); return            |
| async   | no     | f' = new Fut; x ! foo(f'); \           |
|         |        |   f'.chain(null, \lam); return         |
| sync    | yes    | f' = new Fut; x ! foo(f'); \           |
|         |        |   f'' = new Fut(); f'.chain(f'',\lam); \ |
|         |        | return get f''                         |
| sync    | not    | f' = new Fut; x ! foo(f'); \           |
|         |        |   f'.chain(null,\lam); return          |
```
#### 3. `forward PreexistingFuture`
```
      forward (f)
```
where `f` is the preexisting future. The `foward(f)` will be desugared as `foward(f ~~> id)`
```
class Base
  def one(arg : Fut[int]) : int
    forward(arg)
  end
end
```
Another example is to chain on preexisting future.
```
class Base
  def one(arg : Fut[int]) : int
    forward(arg ~~> fun(x:int) => x * 10)
  end
end
```
#### 4. `forward in closure's body`
```
      f ~~> fun(x:int) => forward(g)
```
where `g` is the preexisting future.
```
class Base
  def one(g : Fut[int]) : int
    get( (new Foo) ! foo() ~~> fun(x : int) => forward(g))
  end
end
```
### Examples


PhucVH888:
@supercooldave sorry for losing the comments when re-creating the PR. The last case will be pushed in this PR as well.

supercooldave:
> 1. DONE: forward MethodCall (#682)
> 2. forward FutureChaining
> 3. forward PreexistingFuture
> 4. TODO: forward in closure's body

So this means that 2 and 3 are addressed in this PR, right?

PhucVH888:
Yes, this PR included 2 and 3. Due to your previous comments on 3 that can be implemented by constructing a new AST node which is _not_ parsed. Those commented codes are kept for further discussion.

I intended to create a new PR for number 4, but it would be included in this PR.

src/back/CodeGen/CCodeNames.hs
```diff
@@ -237,6 +237,9 @@ encoreCreateName = Nam "encore_create"
 encoreAllocName :: CCode Name
 encoreAllocName = Nam "encore_alloc"
 
+encoreForwardId :: CCode Name
```
supercooldave: This is no longer used, correct? Please remove.

src/back/CodeGen/Expr.hs
```diff
@@ -1062,11 +1054,41 @@ instance Translatable A.Expr (State Ctx.Context (CCode Lval, CCode Stat)) where
...
+      return (unit, Seq $
+                      [tfuture,
+                       tchain,
+                       (Statement $
+                          Call futureChainActorForward
+                           [AsExpr encoreCtxVar, AsExpr nfuture, ty, AsExpr nchain, AsExpr futVar]
+                       )] ++
+                      [dtraceExit,
+                      Return Skip])
+    else if Ty.isFutureType $ A.getType fchain
```
supercooldave: This test is not required. The type system guarantees that fchain has type Fut[t].

src/back/CodeGen/Expr.hs
```diff
@@ -1062,11 +1054,41 @@ instance Translatable A.Expr (State Ctx.Context (CCode Lval, CCode Stat)) where
...
+          theGet = fromEncoreArgT resultType (Call futureGetActor [encoreCtxVar, nfchain])
+      return $ (Var tmp, Seq $
+                      [tfuture,
+                       tchain,
+                       (Assign (Decl (C.future, Var result))
+                               (Call futureChainActor
+                                 [AsExpr encoreCtxVar, AsExpr nfuture, ty, AsExpr nchain]
+                               )),
+                       Assign (Decl (resultType, Var tmp)) theGet])
+    else
```
supercooldave: The else branch can be removed, due to comment above.

src/opt/Optimizer/Optimizer.hs
```diff
@@ -128,3 +129,28 @@ dropBorrowBlocks = extend dropBorrowBlock
            ,decls = [([VarNoType name], target)]
            ,body}
       dropBorrowBlock e = e
+
+forwardExistingFuture = extend forwardExistingFuture'
+  where
+    forwardExistingFuture' e@(Forward{emeta, forwardExpr=arg@VarAccess{}})=Forward{emeta=emeta', forwardExpr=newExpr}
```
supercooldave: I don't think this is general enough. It should capture every case that is not a method call and not a future chaining. So those cases should be matched first and not transformed, then any other case should be handled by this rule. All those cases will be handled the same.
```
forward(e) ==> forward(e ~~> id)
```

src/opt/Optimizer/Optimizer.hs
```diff
@@ -128,3 +129,28 @@ dropBorrowBlocks = extend dropBorrowBlock
...
+      where
+         forwardExpr = arg
```
supercooldave: This line is redundant. Both `forwardExpr` and `arg` already refer to the same thing.

src/runtime/encore/encore.c
```diff
@@ -443,3 +443,10 @@ void encore_trace_object(pony_ctx_t *ctx, void *p, pony_trace_fn f)
   if (!p) { return; }
   ctx->trace_object(ctx, p, &(pony_type_t){.trace = f}, PONY_TRACE_MUTABLE);
 }
+
+encore_arg_t encore_forward_id(pony_ctx_t **ctx, pony_type_t **t, encore_arg_t value[], void* v) {
```
supercooldave: This function is not used. Eliminate.

src/runtime/encore/encore.h
```diff
@@ -191,4 +191,7 @@ static inline void encore_trace_capability(
   }
 }
 
+/// Identity function for forward
```
supercooldave: Again, not used, remove.

src/types/Typechecker/Typechecker.hs
```diff
@@ -1226,8 +1226,8 @@ instance Checkable Expr where
     doTypecheck forward@(Forward {forwardExpr}) =
         do eExpr <- typecheck forwardExpr
            let ty = AST.getType eExpr
-           unless (isMessageSend forwardExpr) $
-                  pushError eExpr $ ForwardArgumentError
+          --  unless (isMessageSend forwardExpr || isFutureChain forwardExpr) $
+          --         pushError eExpr $ ForwardArgumentError
```
supercooldave: Why are these lines commented out and not removed?

src/types/Typechecker/TypeError.hs
```diff
@@ -448,7 +447,7 @@ instance Show Error where
       | otherwise =
           printf "Cannot compare values across types %s and %s"
                  (show lty) (show rty)
-    show BadSyncCallError = "Synchronous method calls on actors are not allowed (except on the current this)"
+    show BadSyncCallError = "Synchronous method calls on actors is not allowed (except on the current this)"
```
supercooldave: "is" should be "are", which it original was.

src/tests/encore/forward/forwardPreexistingFutureChain.enc
```diff
@@ -0,0 +1,19 @@
...
+active class Base
+  def one(arg : Fut[int]) : int
+    forward(arg ~~> fun (x:int) => x)
+  end
+end
+
+active class Main
+  def main() : unit
+    val arg = (new Foo)!two(42)
+    print("{}\n",get((new Base)!one(arg)))
```
supercooldave: You can use `println` and then you won't need `\n` in the string.

src/back/CodeGen/Expr.hs
```diff
@@ -1062,11 +1054,41 @@ instance Translatable A.Expr (State Ctx.Context (CCode Lval, CCode Stat)) where
...
-    error $ "Expr.hs: Target of forward is not method call or future chain: '" ++
-            show forwardExpr ++ "'"
+  translate A.Forward{A.emeta, A.forwardExpr = fchain@A.FutureChain{A.future, A.chain}} = do
+    (nfuture,tfuture) <- translate future
+    (nchain, tchain)  <- translate chain
+    eCtx <- gets $ Ctx.getExecCtx
+    withForwarding <- gets Ctx.withForwarding
+    let ty = getRuntimeType chain
+        dtraceExit = head (getDtraceExit eCtx)
+    if withForwarding
```
supercooldave: `withForwarding` is misleading name. My understanding is that you translate the code twice, one for when it is called asynchronously, and once for when it is called `synchronously`. The name `withForwarding` should reflect that.

Furthermore, the top-level names in the generated code should reflect that too.

src/opt/Optimizer/Optimizer.hs
supercooldave: Please add a more test general case to show that you can handle this.

supercooldave (commented):
This is an incomplete review. 
I still haven't run the code or applied it to any test cases. 
Please fix the things mentioned.

supercooldave:
The following code does not compile. It should. (Comment already made above, but here is a concrete example.)
```
fun id[t](x: t) : t
  x
end
  
active class Foo
  def foo_arg(arg: Fut[int]) : int
    forward(id(arg))
  end
  def foo() : int
    42
  end
end

active class Main
  def main() : unit
    val foo = (new Foo) ! foo()
    println("{}", get((new Foo)!foo_arg(foo)))
  end
end
```

src/back/CodeGen/Expr.hs
supercooldave: To be honest, I'm not entirely sure that I'm correct. I always viewed methods based on how they were called, rather than what their contents were. This means that whether a method uses forward or not should not affect how it is called. It seems that the implementation is quite different. 

I tried to break the implementation by mixing things up with traits, to trick it into calling the non-forward version of the method when the forward version is expected, but I couldn't do it.

src/back/CodeGen/CCodeNames.hs
PhucVH888: That's true. It is removed.

src/runtime/encore/encore.c
PhucVH888: It's removed.

src/runtime/encore/encore.h
PhucVH888: It's removed.

src/types/Typechecker/Typechecker.hs
PhucVH888: It was for debugging. Removed.

src/types/Typechecker/TypeError.hs
PhucVH888: It has been fixed.

src/tests/encore/forward/forwardPreexistingFutureChain.enc
PhucVH888: It has been replaced by `println(...)`. Thank you.

src/back/CodeGen/Expr.hs
PhucVH888: The `withForwarding` may refer that `forward` is being used, but not explicitly express how the method is called. In this case, I propose to replace `withForwarding` name by `async`.

supercooldave: If it refers only to the fact that `forward` is/may be being used, then perhaps the name doesn't need to be changed.

src/back/CodeGen/Expr.hs
PhucVH888: It is removed.

src/back/CodeGen/Expr.hs
PhucVH888: It is removed.

src/opt/Optimizer/Optimizer.hs
PhucVH888: It is revised.

src/opt/Optimizer/Optimizer.hs
PhucVH888: I have added to more checks. `Forward{MessageSend}` and `Forward{FutureChain}` will be checked first and not be transformed. More test cases have been added.

supercooldave:
> 4. TODO: forward in closure's body

For this case (in the future), there are two variations to consider. When the closure appear with a future chain _and_ for expressions appearing in the body of an `async` (which technically isn't a closure).

supercooldave:
Another test case to consider: `forward(....forward....)`.

PhucVH888:
> Another test case to consider: forward(....forward....).

The test case `forward(...forward...)` has been added.

src/tests/encore/forward/forwardInnerForward.enc
```diff
@@ -0,0 +1,32 @@
...
+  end
+end
+
+active class Foo
+  def foo(b : bool) : int
+    val arg1 = (new Bar) ! bar()
+    val arg2 = (new Base) ! base()
+    forward(if b then
+              forward(arg1)
+              arg1
```
supercooldave: This should be unreachable. Remove.

src/tests/encore/forward/forwardInnerForward.enc
```diff
@@ -0,0 +1,32 @@
...
+            else
+              forward(arg2)
```
supercooldave: How about removing `forward(arg2)`, so that the `if` statement can do something interesting?
What I mean is that you could have one branch of the inner if that uses forward and one that doesn't.

src/opt/Optimizer/Optimizer.hs
```diff
@@ -128,3 +129,30 @@ dropBorrowBlocks = extend dropBorrowBlock
            ,decls = [([VarNoType name], target)]
            ,body}
       dropBorrowBlock e = e
+
+forwardExistingFuture = extend forwardGeneral
```
supercooldave: The name of the visible function is what should have changed. The internal function name can be based on the visible function name. This handles the general case of forward, not just existing futures.

src/types/Typechecker/TypeError.hs
```diff
@@ -712,7 +711,7 @@ instance Show Error where
         printf ("Returned type %s of forward should match with " ++
                "the result type of the containing method %s")
                (show retType) (show ty)
-    show (ForwardArgumentError) = "Forward currently operates on method call"
+    show (ForwardArgumentError) = "Forward currently operates on method call and future chain"
```
supercooldave: This error is surely out of date.

supercooldave:
This definition, adapted from your test case, fails to type check, but I believe it should, simply because `forward` never returns so its type can be anything:
```
active class Foo
  def foo(b : bool) : int
    val arg1 = (new Bar) ! bar()
    val arg2 = (new Base) ! base()
    forward(if b then
              forward(arg1)
            else
              arg2
            end)
  end
end
```
Can't this be fixed by making the type of `forward` bottomType, or does that break things in code generation?

See whether you can fix this one quickly. Otherwise, just use 
```
forward(arg1)
arg1
```
as the `then` clause and file an issue stating what the problem is.

supercooldave (commented):
Some fairly minor comments, mainly about the new test case.

src/opt/Optimizer/Optimizer.hs
PhucVH888: I have fixed it.

src/tests/encore/forward/forwardInnerForward.enc
PhucVH888: Since the `forward`'s type is too restrictive, an extra variable `arg1` is necessary. I have added the comment on the unreachable extra argument `arg1`. A new issue will be created correspondingly.

src/tests/encore/forward/forwardInnerForward.enc
PhucVH888: I have changed it.

src/types/Typechecker/TypeError.hs
PhucVH888: Thanks Dave, it is removed.

PhucVH888:
The issue of restrictive forward's type is described in #792

src/ir/AST/AST.hs
```diff
@@ -748,6 +748,10 @@ isFunctionCall :: Expr -> Bool
 isFunctionCall FunctionCall {} = True
 isFunctionCall _ = False
 
+isFutureChain :: Expr -> Bool
```
supercooldave: This function seems not to be used. 
If that is the case, please remove.

supercooldave (commented):
Some dead code to remove, then you're done.

src/ir/AST/AST.hs
PhucVH888: It's true, this function is not used. Removed.

supercooldave:
I will accept this tomorrow at 10:25, unless others say otherwise.

src/back/CodeGen/Expr.hs
```diff
@@ -1015,18 +1015,10 @@ instance Translatable A.Expr (State Ctx.Context (CCode Lval, CCode Stat)) where
...
-          case eCtx of
-            Ctx.FunctionContext fun ->
-              dtraceFunctionExit (A.functionName fun)
-            Ctx.MethodContext mdecl ->
-              dtraceMethodExit thisVar (A.methodName mdecl)
-            Ctx.ClosureContext clos ->
-              dtraceClosureExit
-            _ -> error "Expr.hs: No context to forward from"
-    if withForwarding
+    let dtraceExit = head (getDtraceExit eCtx)
```
kikofernandez: it seems like this function `getDtraceExit` should always return the first argument since all its uses repeat the pattern `head (getDtraceExit eCtx)`. I think this could be refactored

src/opt/Optimizer/Optimizer.hs
```diff
@@ -128,3 +129,30 @@ dropBorrowBlocks = extend dropBorrowBlock
            ,decls = [([VarNoType name], target)]
            ,body}
       dropBorrowBlock e = e
+
+forwardGeneral = extend forwardGeneral'
+  where
+    forwardGeneral' e@(Forward{forwardExpr=MessageSend{}}) = e
```
kikofernandez, line 135: this case and the one below can be subsumed by the last pattern, `forwardGeneral' e = e`

kikofernandez:
I have provided some refactoring tips that I think should be addressed.
I cannot comment on the logic of the `forward` at this moment

src/opt/Optimizer/Optimizer.hs
supercooldave: @kikofernandez's comment is wrong. These two cases explicitly ignore changes to two `forward` cases. The third case does the work for all other `forward` cases. The final case is a catch all for all other kinds of expressions.

supercooldave (commented):
Comment on kiko's comment.

src/opt/Optimizer/Optimizer.hs
kikofernandez: sorry! you are absolutely right! I misread the code.

kikofernandez:
The comment on `getDtraceExit` should be considered, though

src/back/CodeGen/Expr.hs
PhucVH888: Can you describe more specific?

supercooldave: @kikofernandez kiko means to move `head` into the definition of `getDtraceExit`, rather than have the call to `head` at the 3 or so places where `getDtraceExit` is called. (My interpretation, in any case.)

PhucVH888: I have tried to move `head` into `getDtraceExit` function's body, but it does not work due to type error. @kikofernandez Can you show me how?

PhucVH888: @kikofernandez The code was refactored.

albertnetymk:
AFAICT, `future_chain` is not used externally, so it should be declared as `static` to minimize the API.

Minor:

`actor` occurs in the naming of `future_chain_actor` probably because its first implementation uses `actor` under the hood. Now that another function name is derived from this, I think it makes sense to spend a minute on "fixing" it. Sth like: `future_chain`, `future_chain_forward`, `future_chain_result` for example.

PhucVH888:
@albertnetymk I have fixed it. Thanks for your comment.

PR #785, opened 2017-04-18
Allow any type to instantiate an unsafe parameter

EliasC:
This commit allows using any type where an `unsafe` parameter is
expected, making `unsafe` the escape hatch for parametric
constructs. It is needed for the implementation of `bestow`.


supercooldave:
Does what it claims. Very simple to review.

PR #787, opened 2017-04-20
Added regression test for issue #18

### LINKED ISSUE #18: Closures binding fields broken
Code like the following (in active class `PFilter`)

```
let action = \(i:int) -> { this.filter.cancel(i); (); } in 
  primes.for_all(action)
```

generate C code like the following (manually cleaned, hopefully equivalent)

```
pony_actor_t* this = ((struct ____closure0_env*) _env)->this;
Filter_data* _tmp0 = this->filter;
void* _tmp1 = Filter_cancel(_tmp0, i);
```

which is (correctly) rejected because `this` is a `pony_actor_t` which is opaque, meaning `this->filter` is illegal. 

As far as I can tell, the type of `this` should really be `PFilter_data*`in this case. It seems that the implementation captures `this` rather than `this.filter`, and then attempts to extract `this.filter` later.

supercooldave:
Adds a regression test for (the very old) issue #18.

Works in current development branch.

Fixes #18.


EliasC:
Looks good. Will merge at lunch if no one else objects. I also added "fixes #18", which will make it close the issue automatically when this gets merged!

Issue #404, opened 2016-05-03
Support for compiling standard libraries without importing standard libraries


TobiasWrigstad:
Compiling the _prelude_ bundle(s) (currently just the `String.enc` bundle) is not possible in isolation because the compiler automatically includes the prelude functions in every build.  This makes developing the prelude tricky. 

Fix is probably easy – a new flag and avoiding the include the prelude bundles if the flag is set.

supercooldave:
Fixed in #788.

PR #788, opened 2017-04-20
Fixes issue #404.

supercooldave:
There was a problem compiling in the encore standard library directory -- the compiler would look for modules in the standard directory and in ./ and register an ambiguity. 

This fix checks to see whether ./ is the same as the standard module directory and if so removes it from the directories scanned.

Testing involves going into the standard library directory and compiling `String.enc` with the `-tc` flag and seeing that the previous error is no longer there.

Addresses issue #404.

### DIFF
--- a/src/front/ModuleExpander.hs
+++ b/src/front/ModuleExpander.hs
@@ -17,7 +17,7 @@ import Types(setRefSourceFile, setRefNamespace)
 import SystemUtils
 import Control.Monad
 import Control.Arrow((&&&))
-import System.Directory(doesFileExist)
+import System.Directory(doesFileExist, makeAbsolute)
 import Data.Map.Strict(Map)
 import qualified Data.Map.Strict as Map
 import Data.List
@@ -121,9 +121,14 @@ buildModulePath (NSExplicit ns) =
 findSource :: [FilePath] -> FilePath -> ImportDecl -> IO FilePath
 findSource importDirs sourceDir Import{itarget} = do
   let modulePath = buildModulePath itarget
-      sources = nub $
-                sourceDir </> modulePath :
-                map (</> modulePath) importDirs
+      imports = map (</> modulePath) importDirs
+      sourceModulePath = sourceDir </> modulePath 
+  expandedSourceModulePath <- makeAbsolute $ sourceModulePath
+  let sources = if expandedSourceModulePath `elem` imports then
+                -- if directory of target is in imports, remove it to avoid ambiguous import error
+                  nub $ imports
+                else 
+                  nub $ sourceModulePath : imports                
   candidates <- filterM doesFileExist sources
   case candidates of
     [] -> abort $ "Module " ++ show itarget ++


src/types/Types.hs
```diff
@@ -187,7 +187,7 @@ modeSubtypeOf ty1 ty2 =
   where
     hasMatchingMode modes mode
       | modeIsSharable mode = mode `elem` modes || Sharable `elem` modes
-      | otherwise = mode `elem` modes || Unsafe `elem` modes
+      | otherwise = mode `elem` modes
```
EliasC: Why was this check removed?

src/tests/encore/parametric/unsafe.enc
```diff
@@ -11,9 +10,9 @@ end
 active class Main
   def main(args : [String]) : unit
     val duplicator = new Duplicator[Token]
-    duplicator.f = new Token
-    match duplicator.duplicate() with
-      case (y, y') => println("All your linear are belong to us")
+    var x = new Token
+    match duplicator.duplicate(x) with
+      case (y, y') => println("all your linear are belong to us")
```
EliasC: This commit seems to change things that it shouldn't change...

src/front/ModuleExpander.hs
```diff
@@ -121,9 +121,13 @@ buildModulePath (NSExplicit ns) =
 findSource :: [FilePath] -> FilePath -> ImportDecl -> IO FilePath
 findSource importDirs sourceDir Import{itarget} = do
   let modulePath = buildModulePath itarget
-      sources = nub $
-                sourceDir </> modulePath :
-                map (</> modulePath) importDirs
+      imports = map (</> modulePath) importDirs
+      sourceModulePath = sourceDir </> modulePath 
+  expandedSourceModulePath <- makeAbsolute $ sourceModulePath
+  let sources = if expandedSourceModulePath `elem` imports then -- in std library directory
```
EliasC: What in this line ensures that we are looking at imports of the standard library?

EliasC (changes requested):
This commit seems to revert the changes of previous commits. I suspect there's been a merge issue.

src/types/Types.hs
supercooldave: Oo. I probably forgot to rebase. 
I'll fix. Same with point below.

src/front/ModuleExpander.hs
supercooldave: Technically, it checks that the source directory (of the compilation target) is not among the import directories (including the std libs), and if it is it removes it. 
This is a more robust solution, IMO, but I should document the code to reflect this fact.

supercooldave:
@EliasC I've recommitted. It seems that I'd done a revert to undo some of my changes, but it undid your changed. I checked out afresh and re-added this file. The version now has a comment that hopefully explains the logic.

EliasC (approved):

EliasC:
This looks good! Merging tomorrow.

PR #789, opened 2017-04-28
Disallow silently capturing an active this as a passive reference

### LINKED ISSUE #781: Linear closure accesses shared mutable variable of active object
It seems that a linear closure that accesses the mutable state of an active object can be passed outside of the active object, creating a potential race condition on the mutable state between the closure and the active object. The following example compiles (though it does not exhibit a race condition), when perhaps it shouldn't.

This problem occurs in the `HEAD` of `development`.

```active class Main
  def main() : unit
    val bar = new Bar(10)
    val f = get(bar!clo())
    println(f(12))
  end
end

active class Bar
  var foo : int
  def init(i : int) : unit
    this.foo = i
  end

  def clo() : linear ((int) -> int)
    fun (x : int) : int
      this.foo + x
    end
  end
end
```

EliasC:
Before this commit, in an `active` class, you could use `this` to
access fields or (synchronously) call methods inside a closure.
Because the mode of `this` is `active`, which is safe, the mode of
the closure would allow it to be shared, leading to potential
data-races. This commit changes an `active` `this` to a `local`
variable if it is used for synchronous accesses. This forces the
`local` mode on a closure containing e.g. `this.foo()`.

There are also some improved error messages and a small check for
when a required `var` field is provided as a `val` field.

Fixes #781.

### DIFF
--- a/modules/standard/String.enc
+++ b/modules/standard/String.enc
@@ -11,11 +11,11 @@ BODY
 // This function is called in the very beginning of the program to
 // build an array containing the arguments of the program.
 array_t *_init_argv(pony_ctx_t** ctx, size_t argc, char **argv) {
- array_t *arr = array_mk(ctx, argc, &_enc__read_String_String_type);
+ array_t *arr = array_mk(ctx, argc, &_enc__class_String_String_type);
  for(int i = 0; i < argc; i++) {
-   _enc__read_String_String_t* s =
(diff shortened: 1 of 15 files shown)


supercooldave:
One of the tests fails.

EliasC:
Rebased and repushed!

PR #790, opened 2017-05-02
Add range information to parsed nodes

EliasC:
This commit extends the source position to include the range of
positions of an Encore AST-node. Before, when parsing an expression,
e.g. `(2,3)`, we would only know where the expression began. Now we also
have information on where the expression ends (in this example 5
characters after the beginning).

This is not visible to the programmer in any way, but is integral for
the upcoming editor mode support.

### DIFF
--- a/src/parser/Parser/Parser.hs
+++ b/src/parser/Parser/Parser.hs
@@ -26,7 +26,8 @@ import Control.Arrow (first, (&&&))
 import Identifiers hiding(namespace)
 import Types hiding(refType)
 import AST.AST
-import AST.Meta hiding(Closure, Async, getPos)
+import AST.Meta hiding(Closure, Async, getPos, setEndPos)
+import qualified AST.Meta as Meta(setEndPos)
 
 -- | 'parseEncoreProgram' @path@ @code@ assumes @path@ is the path
 -- to the file being parsed and will produce an AST for @code@,
@@ -169,7 +170,9 @@ blockedConstruct header = do
     constructor <- header
     parseBody constructor
   atLevel indent $ reserved "end"
-  return block
+  end <- getPosition
+  return . setEndPos end $
+    block
 
 -- | These parsers use the lexer above and are the smallest
 -- building blocks of the whole parser.
@@ -469,22 +472,24 @@ program = do
 moduleDecl :: EncParser ModuleDecl
 moduleDecl = option NoModule $
   lineFold $ \sc' -> do
-    modmeta <- meta <$> getPosition
+    modmeta <- meta . newPos <$> getPosition
     reserved "module"
     lookAhead upperChar
     modname <- Name <$> identifier
     modexports <- optional $
                   folded parens sc' ((Name <$> identifier) `sepEndBy` comma)
-    return Module{modmeta
-                 ,modname
-                 ,modexports
-                 }
+    end <- getPosition
+    return . setEndPos end $
+      Module{modmeta
+            ,modname
+            ,modexports
+            }
 
 importdecl :: EncParser ImportDecl
 importdecl =
   lineFold $ \sc' -> do
     indent <- L.indentLevel
-    imeta <- meta <$> getPosition
+    imeta <- meta . newPos <$> getPosition
     reserved "import"
     iqualified <- option False $ reserved "qualified" >> return True
     itarget <- explicitNamespace <$> modulePath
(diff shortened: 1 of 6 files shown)


EliasC:
Most changes are boilerplate changes to the parser to add ending positions of the parsed things.

src/parser/Parser/Parser.hs
```diff
@@ -514,13 +522,13 @@ embedTL = do
   (try (do string "EMBED"
            header <- manyTill anyChar $ try $ do {spaceChar; string "BODY"}
            code <- manyTill anyChar $ try $ do {spaceChar; reserved "END"}
-           return $ EmbedTL (meta pos) header code
+           return $ EmbedTL (meta (newPos pos)) header code
```
kikofernandez, line 525: this is a bit of repetition. Would it work if you:
```haskell
imeta <- meta . newPos <$> getPosition
```

src/parser/Parser/Parser.hs
```diff
@@ -531,7 +539,7 @@ optionalTypeParameters = option [] (brackets $ commaSep1 modedTypeVar)
 
 typedef :: EncParser Typedef
 typedef = do
-  typedefmeta <- meta <$> getPosition
+  typedefmeta <- meta . newPos <$> getPosition
```
kikofernandez, line 542: It seems that this pattern (`meta . newPos`) creates a new `Position` to wrap it inside a `Meta`. Does it make sense to create a function that returns the `Meta` from a `Position`? I think this is less cognitive load, as I don't care too much about how I got the `meta` from a `Position`

src/parser/Parser/Parser.hs
```diff
@@ -579,7 +589,8 @@ localFunction = do
   funIndent <- L.indentLevel
   fun <- funHeaderAndBody
   atLevel funIndent $ reserved "end"
-  return fun
+  returnWithEnd
```
kikofernandez, line 592: I like it more in a single line :)

src/parser/Parser/Parser.hs
```diff
@@ -589,12 +600,12 @@ globalFunction = do
...
   atLevel funIndent $ reserved "end"
-
-  return fun{funlocals}
+  returnWithEnd
+    fun{funlocals}
 
 funHeaderAndBody =
   indentBlock $ do
-    funmeta <- meta <$> getPosition
+    funmeta <- meta . newPos <$> getPosition
```
kikofernandez, line 608: Comment regarding `meta . newPos` applies here also

src/parser/Parser/Parser.hs
```diff
@@ -900,73 +917,89 @@ expression = makeExprParser expr opTable
                   op "/=" DIV_EQUALS]
                 ]
 
+      opRange emeta = do
```
kikofernandez, line 920: the name `opRange` may seem to suggest that it has something to do with `Ranges` in Encore. As I understood it, this operation is used whenever we have an expression in which there may be ambiguity about the pos (column) of the error, e.g. `x ! foo()`, the errors occurs in the `x` or in the `foo`? I don't know yet what better name we can give it but, if the name doesn't change, I think that a comment can help in understanding its _raison d'être_

src/parser/Parser/Parser.hs
```diff
@@ -900,73 +917,89 @@ expression = makeExprParser expr opTable
...
                       name <- Name <$> identifier
                       typeArguments <- option [] (try . brackets $ commaSep typ)
                       args <- parens arguments
-                      let msgSend opt target = MessageSend {emeta = meta pos
-                                                           ,typeArguments
-                                                           ,target
-                                                           ,name
-                                                           ,args}
-                      return $ msgSend False)
+                      emeta' <- opRange emeta
```
kikofernandez, line 969: wouldn't it make sense to abstract the two lines:
```haskell
do emeta <- meta . newPos <$> getPosition
     emeta' <- opRange emeta
```

into a single function that returns the range?

kikofernandez:
I didn't see that this was merged!
In my opinion, this PR should **NOT** have been merged yet! There is quite a bit of refactoring that should have been done

kikofernandez:
now we will have to live with these issues unless someone takes them

src/parser/Parser/Parser.hs
EliasC: :+1:

src/parser/Parser/Parser.hs
EliasC: :+1:

We were doing the translation semi-automatically and were lazy.

src/parser/Parser/Parser.hs
EliasC: It's a question of where to put the complexity. In the parser, we get a `SourcePos` to create a `Meta` from. In the rest of the compiler, we create new `Meta`s by getting the `Position`s from other `Meta`s. We went with this solution since it kept the complexity local to a single module. An alternative would be to expose two constructors for `Meta`s, `meta` and `metaRaw` or something, but we would still need to use different functions in different places. Is this better in your opinion?

src/parser/Parser/Parser.hs
EliasC: :+1:

EliasC:
Addressed @kikofernandez' comments in a new PR: #796

Issue #759, opened 2017-03-29
Gets C warning when for loop has non-unit type


albertnetymk:
```
class Main
  def f() : int
    1
  end

  def main() : unit
    for i <- [0 .. 10] do
      this.f()
    end
  end
end
```

Compiling emits C warning `incompatible integer to pointer conversion assigning to 'void *' from 'int64_t' (aka 'long')`.

PR #794, opened 2017-05-03
Adding Pedestrian simulation stress test

helanhalvan:
This PR adds my pedestrian simulation code into the configuration with other stress tests. As I don't really know your needs, I have made it easy to re-configure. All the code is in /pedsim/Ped_util, while instance data files and startup files are in /pedsim. 

For configuration options, see src/tests/stress/pedsim/README.md

My code also emit c compile warning, described here: https://github.com/parapluu/encore/issues/759


src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,123 @@
...
+    this.x = x
+    this.y = y
+  end
+  def pos() : (int, int)
+    (this.x, this.y)
+  end
+  def ttldec() : unit
+    this.ttl=this.ttl-1
+  end
+{--
```
kikofernandez: if this is not needed, it should be gone

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,123 @@
...
+          ret(2) = desired_x
+          ret(3) = y1
+          ret(4) = x1
+          ret(5) = desired_y
+        end
+      end
+      ()
+    end
+  end
+  def move(in_pos : (int, int)) : unit
```
kikofernandez: you have two methods, `move` and `move_int`. they seem to do pretty much the same thing, except that they can receive a tuple or two numbers. Why do you need both functions?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+  int i = *(int*)&x;
+  i = 0x5f3759df - (i >> 1);
+  x = *(float*)&i;
+  #{ret}= x*(1.5f - xhalf*x*x);
+  }
+  END
+  ret
+end
+fun find_extreme(agents : [Agent]) : (int, int, int, int)
+  var extreme = (-(10000), -(10000), 10000, 10000)
```
kikofernandez: these are magical numbers. Where do they come from? Wouldn't it be better if they have some doc regarding the number ±1000?

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
helanhalvan: I used both in earlier versions and forgot to remove them. I'll get rid of `move`.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+    end
+  end
+  match extreme with
+    case (xmax, ymax, xmin, ymin) =>
+      (xmax + 10, ymax + 10, xmin - 10, ymin - 10)
+    end
+
+  end
+end
+fun extreme_check(a : (int, int), b : (int, int, int, int)) : (int, int, int, int)
```
kikofernandez: you should try to be consistent when naming things, e.g. `fun invSqrt(input : int) : real` vs `fun extreme_check(a : (int, int), b : (int, int, int, int)) : (int, int, int, int)`. Choose one, both do not make sense (camel case or snake case)

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+  var list = new [(int, int)](len)
+  var i = 0
+  for a <- listlist do
+    for b <- a do
+      list(i) = b
+      i = i + 1
+    end
+  end
+  unless i == len then
+    print("BAD MERGE\n")
```
kikofernandez: should this be an error?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+      nr
+    end
+    case _ =>
+      print("or0\n")
+      0
+    end
+
+  end
+end
+fun string_or_empty(i : Maybe[String]) : String
```
kikofernandez: your `string_or_empty` is more likely a `maybeToString` that takes `""` for the value `Nothing`

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+    end
+    case _ =>
+      print("orEmpty\n")
+      ""
+    end
+
+  end
+end
+
+fun rot(nr : real) : real
```
kikofernandez: I think that this is `sqrt`

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+  nr
+end
+fun abs(a : int) : int
+  if a > 0 then
+    a
+  else
+    a * -(1)
+  end
+end
+fun man_distance(a : (int, int), b : (int, int)) : int
```
kikofernandez: same in this case, are `man_distance` and `man_distance_int` both needed?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: I'll just re-write it to not contain magic numbers, it's just a lazy way to initialize the extreme value tuple.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+  match (a, b) with
+    case ((x1, y1), (x2, y2)) =>
+      abs(x1 - x2) + abs(y1 - y2)
+    end
+    case _ =>
+      print("THIS WILL NEVER HAPPEN!")
+      1 / 0
```
kikofernandez: use a proper exit, not this trick

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+  i
+end
+fun floor(x : real) : int
+  val i = 0
+  EMBED (unit)
+    { #{i}=floor(#{x}); }
+  END
+  i
+end
+fun distance(a : (int, int), b : (int, int)) : real
```
kikofernandez: another distance. Are all these 4 different versions of distant needed?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,315 @@
...
+  match (a, b) with
+    case ((x1, y1), (x2, y2)) =>
+      rot(x1 - x2 * x1 - x2 + y1 - y2 * y1 - y2)
+    end
+    case _ =>
+      print("THIS WILL NEVER HAPPEN!")
+      1 / 0
```
kikofernandez: throw an error

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
```diff
@@ -0,0 +1,154 @@
...
+    this.min_y = min_y
+    this.top_right=Nothing
+    this.top_left=Nothing
+    this.bottom_right=Nothing
+    this.bottom_left=Nothing
+    this.size = 0
+    this.agent = new [int](0)
+  end
+  def add(x : int, y : int) : unit
+    --if x <= this.max_x && x >= this.min_x && y <= this.max_y && y >= this.min_y then
```
kikofernandez: remove line unless it's needed

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
```diff
@@ -0,0 +1,154 @@
...
+                this.size = 1
+            else
+              val dx = (this.max_x - this.min_x) / 2
+              val dy = (this.max_y - this.min_y) / 2
+              this.top_right = Just(new Quad(this.max_x, this.max_y, this.min_x + dx + 1, this.min_y + dy + 1))
+              this.top_left = Just(new Quad(this.min_x + dx, this.max_y, this.min_x, this.min_y + dy + 1))
+              this.bottom_right = Just(new Quad(this.max_x, this.min_y + dy, this.min_x + dx + 1, this.min_y))
+              this.bottom_left = Just(new Quad(this.min_x + dx, this.min_y + dy, this.min_x, this.min_y))
+              this.size = 2
+              --print("deeper {} {} {} {} {} {}\n", this.max_x, this.max_y, x, y,this.agent(0),this.agent(1))
```
kikofernandez: to be removed

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
```diff
@@ -0,0 +1,154 @@
...
+        this.ymin = ymin
+      end
+    end
+  end
+  def get_val(x : int, y : int) : bool
+    this.tree.isin(x, y)
+  end
+  def set(x : int, y : int) : unit
+    if x > this.xmax || y > this.ymax || y < this.ymin || x < this.xmin then
+      print("TRYING TO STORE VALUE OUTSIDE OF THIS TREE\n")
```
kikofernandez: is this an error?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
+module Regions
+import Agent_passive
+import Global_funs
+import Quad_tree
+
+fun regions(var agents : [Agent], boxes : int) : unit
+  var box = new Tiling_box(agents, boxes)
+  box!move(10000)
```
kikofernandez: where does the 10000 come from?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+end
+fun link(a : Box, b : Box) : bool
+  val aa = a!link(b)
+  val bb = b!link(a)
+  val aaa = get(aa)
+  val bbb = get(bb)
```
kikofernandez, line 14: you are getting on `bb` but not using the result. Is this for synchronisation purposes?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+            end
+          end
+        end
+      end
+
+    end
+  end
+
+  def move(i : int) : unit --,sdl : SDL_Main
+    --print("move\n")
```
kikofernandez: remove print

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+    for id <- [0..|this.boxes| - 1] do
+      (this.futures)(id) = ((this.boxes)(id)) ! move()
+    end
+    --print("waiting\n")
```
kikofernandez: remove

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+    for f <- this.futures do
+      get(f)
+    end
+    if i > 1 then
+      this!move(i - 1)
+    else
+      print("DONE")--printtime(clock() - start_time)
```
kikofernandez: is this print necessary? If performed many times, it may affect the performance

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+    end
+  end
+  def agents() : [(int, int)]
+    val temp = new [[(int, int)]](|this.boxes|)
+    for i <- [0..|this.boxes| - 1] do
+      temp(i) = get((this.boxes(i)) ! agents())
+    end
+    --print("FLATTENING!\n")
```
kikofernandez: remove comments

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+                    case Just(thing) =>
+                     cur=thing.next
+                     ret(i) = match thing.a?.pos() with
+                                case Just(something) => something
+                              end
+                    end
+                end
+                i = i + 1
+              end
+              --print("bas\n")
```
kikofernandez: remove

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: It's an old debug check. I'll just remove it.

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+      if this.matrix.get_val(x, y) == true then
+        a --consume(a)
+      else
+        a?.move_int(x, y)
+        this.add_internal(a,x,y)
+        Nothing
+      end
+    end
+  end
+  def add_internal(a : Maybe[Agent], x : int, y : int) : bool
```
kikofernandez: is this one somehow a `private` method? It seems that `add` is the public interface.

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+    if this.matrix.get_val(x,y) == true then
+      print("AGENTS IN TOP OF eachother\n")
```
kikofernandez: is this needed?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+      while true do
+        ()
```
kikofernandez: how do you exit from this loop?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+        this.agents = this.newcommers
+      else
+        this.agents = this.newcommers
+      end
+      this.newcommers = Nothing
+      this.last = Nothing
+    end
+  end
+  --private
+  def is_something(a:Maybe[Item]) : bool
```
kikofernandez: there is a `private` keyword that makes the method `private`

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+     match a with
+        case Nothing => false
+        case unused  => true
+     end
+  end
+    -- ?. operator call fun on maybe type
+  def move() : bool
+    --print("{}\n",this.ttl)
```
kikofernandez: remove?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,480 @@
...
+    end
+  end
+  --private
+  def is_something(a:Maybe[Item]) : bool
+     match a with
+        case Nothing => false
+        case unused  => true
+     end
+  end
+    -- ?. operator call fun on maybe type
```
kikofernandez: what is this comment?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: I think that name makes sense, it takes a Maybe[String] and makes it a String or an empty String.

src/tests/stress/pedsim/parallel_big.enc
```diff
@@ -0,0 +1,30 @@
+import Ped_util.Global_funs
```
kikofernandez: please do not include binary files, i.e. remove the `src/tests/stress/pedsim/parallel_big` file

src/tests/stress/pedsim/parallel_big.enc
```diff
@@ -0,0 +1,30 @@
...
+  def main(args : [String]) : unit
+    let
+      agents = parse_file("16000_scenario.xml") --args(1)
+      boxes = 9
+    in
+     regions(agents,boxes)
+    end
+  end
+end
+{--
```
kikofernandez: remove this comment if this is not needed

src/tests/stress/pedsim/parallel_small.enc
```diff
@@ -0,0 +1,30 @@
...
+  def main(args : [String]) : unit
+    let
+      agents = parse_file("4000_scenario.xml") --args(1)
+      boxes = 9
+    in
+     regions(agents,boxes)
+    end
+  end
+end
+{--
```
kikofernandez: remove this comment / class if this is not needed

src/tests/stress/pedsim/serial_big.enc
```diff
@@ -0,0 +1,41 @@
...
+                    5
+                  end
+                end
+              end--}
+    in
+     regions(agents,boxes)
+    end
+  end
+end
+{--
```
kikofernandez: remove

src/tests/stress/pedsim/Ped_util/Global_funs.enc
kikofernandez: if you need magic numbers, fine as long as they have a reasonable explanation and they are in some place that's accessible. Putting these numbers for no apparent reason is a code smell

kikofernandez:
Before you refactor your code, be careful:
```
TOTAL TESTS FAILED: 3
 - test stress/pedsim/parallel_big failed
 - test stress/pedsim/serial_big failed
 - test stress/pedsim/parallel_small failed
```

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: What is the proper exit in Encore?

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: That's part of the problem spec, 10000 time steps. I can lift it into a the test start if you like.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
kikofernandez: `abort("Text")`

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: It's another setup sanity check, that method is suposed to return `aaa && bbb`.

albertnetymk:
Unless the author plans to clean up the history a bit, I vote for squashing on merging this.

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,92 @@
...
+      y1 = this.y
+    in
+      var x2 = (this.targets_x)(this.i)
+      var y2 = (this.targets_y)(this.i)
+      var xdiff = x1 - x2
+      var ydiff = y1 - y2
+      var lenSquare = xdiff * xdiff + ydiff * ydiff
+      var invlen = inv_sqrt(lenSquare)
+      if lenSquare < (this.targets_size(this.i) * this.targets_size(this.i)) then
+        this.i = this.i + 1
```
kikofernandez: I don't understand why this is done. could you write a comment about it?

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,92 @@
...
+      end
+      if |this.targets_size| == this.i then
+        this.i = 0
```
kikofernandez: I don't understand why this is done. could you write a comment about it?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+end
+fun extreme_check(a : (int, int), b : (int, int, int, int)) : (int, int, int, int)
+  match (a, b) with
+    case ((x, y), (xmax, ymax, xmin, ymin)) =>
+      (max(x, xmax), max(y, ymax), min(x, xmin), min(y, ymin))
+    end
+  end
+end
+
+fun copy_ar(arr : [int]) : [int]
```
kikofernandez, line 38: couldn't this function and `fun copy_tr(arr : [(int, int)]) : [(int, int)]` be done using a parametric function?
If you look at them, they are the exact same code

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+    b
+  end
+end
+fun yolo_int(i : Maybe[int]) : int
+  match i with
+    case Just(nr) =>
+      nr
+    end
+    case _ =>
+      print("or0\n")
```
kikofernandez: does this need to be said? if so, it should go into the `abort` statement

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+
+  end
+end
+fun yolo_string(i : Maybe[String]) : String
+  match i with
+    case Just(nr) =>
+      nr
+    end
+    case _ =>
+      print("orEmpty\n")
```
kikofernandez: same, should this be in the `abort` instead?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+    end
+    case _ =>
+      print("or0\n")
+      abort("parsing error")
+      0
+    end
+
+  end
+end
+fun yolo_string(i : Maybe[String]) : String
```
kikofernandez, line 97: couldn't this function and the previous one (`fun yolo_int(i : Maybe[int]) : int`) be done with a single parametric function?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+                         end
+
+                       end
+        waypoints_distance(i) = yolo_int(a.attribute_value("r").to_int())
+        i = i + 1
+      end
+      var nr_agents = 0
+      for a <- file.children_named("agent") do
+        nr_agents = nr_agents + match a.attribute_value("n").to_int() with
+                                  case Just(nr) =>
```
kikofernandez: isn't this `match` the same as your `yolo` stuff?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+          var dy_max = yolo_int(a.attribute_value("dy").to_int()) / 2
+          var agent_waypoints = new [(int, int)](|a.children_named("addway")|)
+          var dx = x - dx_max
+          var dy = y - dy_max
+          var j = 0
+          do
+            for b <- a.children_named("addway") do
+              agent_waypoints(j) = waypoints(yolo_int(yolo_string(b.attribute_value("id").substring(1, 2)).to_int()) - 1)
+              unless |agent_waypoints| > 1 then
+                print("THIS SHOULD NEVER HAPPEN |new_agent.waypoint|={}\n", |agent_waypoints|)
```
kikofernandez, line 188: shouldn't this be in the `abort`?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+                abort("parsing error")
+              end
+              j = j + 1
+            end
+            for unused <- [0..max-1] do
+              if dx > x + dx_max then
+                print("out of space\n")
+                abort("parsing error")
```
kikofernandez: same, in the `abort`?

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
kikofernandez: remove this lines

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
kikofernandez: remove this comment if not needed

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
```diff
@@ -0,0 +1,154 @@
...
+      false
+    end
+  end
+  def recur(x : int, y : int) : Quad
+    val top = y > (((this.max_y - this.min_y) / 2) + this.min_y)
+    val right = x > (((this.max_x - this.min_x) / 2) + this.min_x)
+    if top then
+      if right then
+        match (this.top_right) with
+            case Just(qq) => qq
```
kikofernandez, line 99: what happens if you get a `Nothing`? same in the cases below

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
kikofernandez: is this an error?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+            (this.boxes)(i) = new Box((xmax, ymax), (x, y))
+            win = win && link((this.boxes)(i), (this.boxes)(i - n))
+            win = win && link((this.boxes)(i), (this.boxes)(i - 1))
+            win = win && link((this.boxes)(i), (this.boxes)(i - n - 1))
+        else
+            this.boxes(0) = new Box((xmax,ymax), (xmin,ymin))
+        end
+        for a <- agents do
+          for b <- this.boxes do
+            if get(b!add(a)) then
```
kikofernandez: this serialises your code. is this side-effect really intended?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+    for id <- [0..|this.boxes| - 1] do
+      (this.futures)(id) = ((this.boxes)(id)) ! move()
+    end
+    for f <- this.futures do
+      get(f)
+    end
+    if i > 1 then
+      this!move(i - 1)
+    else
+      print("DONE")
```
kikofernandez, line 99: should this be removed?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+      end
+    end
+  end
+
+  def move(i : int) : unit
+    for id <- [0..|this.boxes| - 1] do
+      (this.futures)(id) = ((this.boxes)(id)) ! move()
+    end
+    for f <- this.futures do
+      get(f)
```
kikofernandez, line 94: this serialises the whole execution and may get the actor to be schedule and unscheduled many times. not the most performance design decision

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+      print("DONE")
+    end
+  end
+  -- used by ui, if a ui is used
+  def agents() : [(int, int)]
+    val temp = new [[(int, int)]](|this.boxes|)
+    for i <- [0..|this.boxes| - 1] do
+      temp(i) = get((this.boxes(i)) ! agents())
+    end
+    var a=flatten(temp)
```
kikofernandez: you can just return `flatten(temp)`, no need to do the `a` at the end, isn't it?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+        this.ymax = y
+      end
+    end
+    match in_min with
+      case (x, y) =>
+        this.xmin = x
+        this.ymin = y
+      end
+    end
+    this.ttl = 10001
```
kikofernandez: why this magic number?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+      if this.matrix.get_val(x, y) == true then
+        Just(a)
+      else
+        a.move_int(x, y)
+        this.add_internal(a,x,y)
+        Nothing
+      end
+    end
+  end
+  def add_internal(a : Agent, x : int, y : int) : bool
```
kikofernandez: this function seems to be `private`

kikofernandez: please look out for others

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+      true
+    else
+      this.matrix.set(x,y)
+      if this.newcommers == Nothing then
+        this.newcommers = Just(new Item(a))
+      else
+        if this.last == Nothing then
+          this.last = Just(new Item(a))
+          match this.newcommers with
+              case Just(thing) => thing.append(this.last)
```
kikofernandez: could you give a more descriptive name to `thing`? I have no clue of what is represents

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+      else
+        a.move_int(x, y)
+        this.add_internal(a,x,y)
+        Nothing
+      end
+    end
+  end
+  def add_internal(a : Agent, x : int, y : int) : bool
+    if this.matrix.get_val(x,y) == true then
+      print("AGENTS IN TOP OF eachother\n")
```
kikofernandez: needed?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+      while true do
```
kikofernandez: how do you break from this?

src/tests/stress/pedsim/Ped_util/Regions.enc
kikofernandez: this seems to always return `true`. why?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+  end
+  def move() : bool
+    this.ttl = this.ttl - 1
+    this.merge()
+    var parent = this.agents
+    var wrapped_a = this.agents
+    var desired = new [int](6)
+    while this.is_something(wrapped_a) do
+      var a = match wrapped_a with
+                case Just(value) => value
```
kikofernandez: what happens when you get `Nothing`?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+                              this.matrix.remove(x2, y2)
+                              moved = true
+                              if first then
+                                 this.agents = a.next
+                              end
+                        end
+                    end
+                end
+              end
+              i = i + 2
```
kikofernandez: why `i+2`? can you add a comment?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+          end
+        end
+      end
+      wrapped_a = a.next
+    end
+    this.moving = false
+    true
```
kikofernandez: this method seems to always return `true`. Your code here is unreadable, you should consider refactoring

src/tests/stress/pedsim/Ped_util/XML.enc
```diff
@@ -0,0 +1,230 @@
...
+                  Nothing
+                end
+
+              end
+      EMBED (unit)
+        return #{r};
+      END
+    end
+    var start = s.find("<")
+    var finnish = 0
```
kikofernandez: mispelling

src/tests/stress/pedsim/Ped_util/XML.enc
```diff
@@ -0,0 +1,230 @@
...
+
+                end
+      start = s.find_from("<", finnish)
+    end
+    node.name = cont(0)
+    this.ext_atribs(node, node.name)
+    if this.pair(node.name, cont(|cont| - 1)) then
+      node.comments = new [String](|cont|)
+      var ci = 0
+      var babies = new [String](|cont|)
```
kikofernandez, line 96: what is a `baby`?

kikofernandez: maybe better naming

src/tests/stress/pedsim/Ped_util/XML.enc
```diff
@@ -0,0 +1,230 @@
...
+            ()
+          end
+          case Nothing =>
+            bad = true
+            ()
+          end
+
+        end
+      end
+      if bad == false then
```
kikofernandez: what does it mean to be `bad`? use a more descriptive variable name please

kikofernandez (changes requested):
please consider the points that I have mentioned here. Do not forget to reply to all of them.

PhucVH888:
How can we find the description of the Ped Simulation?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+  for a <- agents do
+    i = i + 1
+    extreme = extreme_check(a.pos(), extreme)
+    for index <- [0..|a.targets_x| - 1] do
+      extreme = extreme_check(((a.targets_x)(index), (a.targets_y)(index)), extreme)
+    end
+  end
+  match extreme with
+    case (xmax, ymax, xmin, ymin) =>
+      (xmax, ymax, xmin, ymin)
```
PhucVH888: Why is there only one case?

src/tests/stress/pedsim/Ped_util/IO.enc
```diff
@@ -0,0 +1,74 @@
...
+                    exit(1);
+                  };
+                  file;
+
+                END
+  end
+  def close() : unit
+    var f = this.file
+    this.file = EMBED (EMBED FILE* END)
+                  if (!#{f}) {
```
PhucVH888, line 36: Minor: please check the indentation and style.

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,92 @@
+module Agent_passive
+import Global_funs
+unsafe class Agent
```
PhucVH888: Could you briefly add some comments for the class `Agent` to describe what it does?

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,92 @@
...
+    end
+    this.hash_code = in_id
+    this.ttl = ttl
+    ()
+  end
+  def next(ret : [int]) : unit
+    let
+      x1 = this.x
+      y1 = this.y
+    in
```
PhucVH888: Why `let..in` is necessary here?

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,92 @@
...
+      xdiff = x1 - x2
+      ydiff = y1 - y2
+      lenSquare = xdiff * xdiff + ydiff * ydiff
+      invlen = inv_sqrt(lenSquare)
+      let
+        xoff = xdiff * invlen
+        yoff = ydiff * invlen
+        desired_x = x1 + round(xoff * -(1))
+        desired_y = y1 + round(yoff * -(1))
+      in
```
PhucVH888: Similar question as above, why `let..in` is necessary here?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+    end
+  end
+end
```
PhucVH888: What is the expected result of `find_extreme()` in case the `extreme` does not match with the only pattern?

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+fun parse_file(fname : String) : [Agent]
+  var sen = new XML_lib()
+  var file = new XML_node()
+  do
+    match sen.file_to_xml(fname) with
+      case Just(xml) =>
+        file = xml
+      end
+      case Nothing =>
+        print("\n\nBADXML!\n")
```
PhucVH888: Please revise the error message.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
```diff
@@ -0,0 +1,230 @@
...
+                else
+                  dy = dy + 1
+                end
+              end
+              agents(i) = new Agent((dx, dy), copy_tr(agent_waypoints), copy_ar(waypoints_distance), i, 10000)
+              i = i + 1
+            end
+          end
+        end
+        agents --consume
```
PhucVH888, line 209: Why the comment is necessary here?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+  end
+end
+local class Item : Id
+  var a : Agent
+  var next : Maybe[Item]
+  var prev : Maybe[Item]
+  def init(var a : Agent) : unit
+    this.a = a --consume(a)
```
PhucVH888, line 434: Again, why the comment is needed here? In which case `consume(a)` would be used?

src/tests/stress/pedsim/README.md
```diff
@@ -0,0 +1,5 @@
+This folder contains the implementation of a pedestrian simulation in Encore. The simulation uses input XML files, which by conventions have the name X_scenario.xml, where X is the number of agents they contain. The various test instances uses different files split into different numbers of regions.
+
+Note: The number of parallel regions/active-object workers is the input number squared. So regions(agents,3) will make 9 workers.
+
```
PhucVH888: The description of the PedSim could be briefly summarised here, IMO. Or the link the the document.

src/tests/stress/pedsim/parallel_small.enc
```diff
@@ -0,0 +1,16 @@
+import Ped_util.Global_funs
+import Ped_util.Regions
+import Ped_util.Quad_tree
+
+
+class Main
+  def main(args : [String]) : unit
+    let
+      agents = parse_file("400_scenario.xml")
+      boxes = 9
```
PhucVH888, line 10: What is the range of `boxes`?

src/tests/stress/pedsim/parallel_small.enc
```diff
@@ -0,0 +1,16 @@
...
+      steps = 10000
```
PhucVH888, line 11: And range of steps?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -0,0 +1,462 @@
...
+                    this.matrix.set(x, y)
+                    a.a.move_int(x, y)
+                    moved = true
+                  end
+                end
+                case Just(real_thing) =>
+                    x2 = ax
+                    y2 = ay
+                    val first= (wrapped_a == this.agents)
+                    var nejbor = real_thing ! external_move( a.a, x, y) --consume
```
PhucVH888, line 403: Again, when `--consume` could be used?

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
```diff
@@ -0,0 +1,92 @@
...
+  def next(ret : [int]) : unit
+    let
+      x1 = this.x
+      y1 = this.y
+    in
+      var x2 = (this.targets_x)(this.i)
+      var y2 = (this.targets_y)(this.i)
+      var xdiff = x1 - x2
+      var ydiff = y1 - y2
+      var lenSquare = xdiff * xdiff + ydiff * ydiff
```
PhucVH888: The variables' names (`lenSquare`, `xdiff`, and many more) should uniformly follow the either snake case or camel case, but both styles shouldn't be mixed up.

PhucVH888 (changes requested):

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
helanhalvan: yes, changed to an abort

supercooldave: Can the message be not in all caps?

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
helanhalvan: It's the waypoint reached check, added comment about it.

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
helanhalvan: This if ensures that waypoints are reused is they run out. Added comment

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: In my experience, attempts at using parametric and/or polymorphic constructs have resulted in a lot of crashes and no benefit. Therefore, I'll not use any parametric method/functions or polymorphism.

This might no longer be a problem but I don't care, I got burned to many times.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: moved to abort statement

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: See earlier comment about parametric functions.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: moved to abort

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: yes, replaced by yolo_int

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: abort does not support parameters, so the int needs to be printer separately anyway.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: fixed

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
helanhalvan: fixed

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
helanhalvan: fixed

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
helanhalvan: You can't, recur only gets called after the node is known to not be a leaf. If you did tho, you would crash, which is what we want in that case.

src/tests/stress/pedsim/Ped_util/Quad_tree.enc
helanhalvan: yes, it's an abort now.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: This is method is only called in setup, which is NOT optimised at all, and not measured in performance measurements. So yea, it's serial.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: No

First of all, you should know that all the boxes can work on fulfilling their futures in parallel, which means that the execution is not serialised. In this active object the execution is serialised, but that's the only reason active objects are thread safe at all.

I have explained this design in my report, and in several of our meetings earlier, as well as shown that this design indeed is capable parallel execution.

Secondly, it's also a requirement of the problem specification that all active objects synchronise between time steps, and they do this by fulfilling futures.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: No.

The program prints DONE when it's done. `move(i)` moves all agents within it `i` time steps.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: removed it

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: removed it, it's the number of time steps +1, used for detecting agents that otherwise might move twice per time step.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: fixed

helanhalvan: changed to unit

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: another setup sainity check, is an abort now

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: you don't, replaced with abort

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: No, replaced with yolo_item

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: You don't, `while this.is_something(wrapped_a) do` might hint that quite clearly.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: We are iterating a [int] that contain 2D coordinates stored [x1,y1,x2,y2,x3,y3], added comment.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: Changed to unit. I won't re-factor, not worth the effort as verifying that this method is correct is difficult, and I have tested it as it looks now.

src/tests/stress/pedsim/Ped_util/XML.enc
helanhalvan: fixed

src/tests/stress/pedsim/Ped_util/XML.enc
helanhalvan: An XML child, changed name.

src/tests/stress/pedsim/Ped_util/XML.enc
helanhalvan: changed name to error

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: removed

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: it's statically known to not happen, also removed

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: changed to `abort("XMLPARSEING ERROR\n")`

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: agent is supposed to be linear, but can't be due to limitations of the current type system. In order to make this easier too change later, I added `-- consume` where it actually should say `consume(...)`

src/tests/stress/pedsim/Ped_util/Global_funs.enc
supercooldave: Again, why all caps?
XML Parsing

src/tests/stress/pedsim/Ped_util/IO.enc
helanhalvan: Added comment about that file and `XML.enc`, both should ideally be replaced with standard library stuff when that is available.

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
helanhalvan: Added tiny comment at the top of the file, also comment before `next` as that's the only method with more then 5 lines outside of the constructor.

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
helanhalvan: removed

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
helanhalvan: I'll try to make it all snake_case.

src/tests/stress/pedsim/Ped_util/Agent_passive.enc
helanhalvan: have the same effect as a lot of val's, and it's a part of the syntax, so I use it.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: see earlier comment

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: It can't, see earlier comment

src/tests/stress/pedsim/README.md
helanhalvan: I'll add a link to the report when it's published somewhere. That's a good idea. Need the report to be hosted with other done bachelor thesis's reports first.

src/tests/stress/pedsim/parallel_small.enc
helanhalvan: Any integer greater then or equal to 1, changed input type to `uint`.

src/tests/stress/pedsim/parallel_small.enc
helanhalvan: uint, changed input type to enforce that.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
helanhalvan: Why is all caps for error messages a problem?

src/tests/stress/pedsim/Ped_util/Regions.enc
```diff
@@ -332,19 +323,22 @@ active class Box
         case unused  => true
      end
   end
-  def move() : bool
+  def private yolo_item(a:Maybe[Item]) : Item
```
kikofernandez: this function is named `yolo_item` but what it's doing is extracting whatever is in an option type.
Doesn't it make more sense to write the following function:
```python
fun fromJust[a](x: Maybe[a]): a
    match x with
        case Just(value) => value
        case Nothing     => abort("Error: XXXX") -- or something
    end
end
```

kikofernandez (commented):
I have one more minor refactoring. In general, it's better to give names that tell you what the function does and/or uses a common and well understood name. In this case, I had no idea of what does the `yolo_item(a: Maybe[Item])` method does. I thought it was setter but it's actually an extractor of the value of the option type. I believe it makes more sense as a global function that can be re-used regardless of the object.

src/tests/stress/pedsim/Ped_util/Regions.enc
helanhalvan: The name yolo is taken from Tobias that suggested to add a method with that name for doing exactly what I am doing in these methods (`yolo_int`, `yolo_string`, `yolo_item`). The reason `yolo_item` is a private method is that `Item` is a local class. And again, I don't trust parametric and/or polymorphic constructs in this language, and will therefore not use them.

src/tests/stress/pedsim/Ped_util/Global_funs.enc
supercooldave: It's completely unnecessary to have all caps. Only crazy people on the internet use all caps. Are you a crazy person on the internet? ARE YOU? (See how crazy it looks.)

helanhalvan: If the XML parsing breaks, I am. Fixed it anyway, just to make you less sad.

src/tests/stress/pedsim/Ped_util/Regions.enc
supercooldave: I wonder whether this can be implemented more efficiently (this is in part a note to @kikofernandez etc). For example, if all the futures are chained together and wrapped with a single `get`, then that would involve only a single `get`, which is the most expensive operation here (except when a future is fulfilled already). With three futures, the code would look like:
`get (f1 ~~> (fun (x : t) f2 ~~> (fun (x : t) f3 ~~> fun (x : t) ())))` (plus tweaking to get types etc right).
This is probably far from optimal, given that creating a future is also expensive and this would create 3 more.

helanhalvan: You will all be free to experiment with all that when this is merged. I don't think a giant future chain is a good idea tho. Maybe some custom method (parallel combinator?) can do stuff like `[Fut T] -> Fut T` effectively. I think this kind of programming pattern is quite common.

Another option is to kill all the futures, and use reply messages instead, but I don't think that's the direction Encore should go, as futures are nice.

kikofernandez (approved):

kikofernandez:
I will merge this PR tomorrow at 15.00 unless someone has any further comment that should be taken into consideration.

kikofernandez:
@PhucVH888 it would be good if you agree and update your review status. Otherwise, please provide further comments.

PhucVH888 (approved):

PhucVH888:
@kikofernandez I think it's good enough, you can merge it if there's not any other comments.

PR #795, opened 2017-05-03
Fix bug where type variable names could clash

### LINKED ISSUE #791: Shared polymorphic type name emits C error
If you name a polymorphic type `t` in a class and have a trait with a method that uses the same name `t` for a polymorphic type, you get a C compilation error:
```
trait Bar
  def bar[t](id : t) : t
    id
  end
end

local class Foo[t] : Bar
end

active class Main
  def main() : unit
    ()
  end
end
```

The faulty generated C code:
```
1.  encore_arg_t _enc__method__Main_Foo_bar(pony_ctx_t** _ctx, _enc__local__Main_Foo_t* _this, 2 2.  pony_type_t** runtimeType, encore_arg_t _enc__arg_id)
3.  {
4.   ENC_DTRACE3(METHOD_ENTRY, (uintptr_t)*_ctx, (uintptr_t)_this, "bar");
5.   pony_type_t* _enc__type_t = (runtimeType[0]);
6.   pony_type_t* _enc__type_t = (*((_enc__local__Main_Foo_t*) _this))._enc__type_t;
7.   ENC_DTRACE3(METHOD_EXIT, (uintptr_t)*_ctx, (uintptr_t)_this, "bar");
8.   return ((encore_arg_t) _enc__arg_id);
9.  }
```
As you can see, on line 6 the variable `_enc__type_t` is redefined from the previous line which emits the error.

EliasC:
This commit fixes a bug where the C names of type variables could clash
if a parametric class (`class C[t]`) included a trait with a polymorphic
method (`def m[t](x : t) : t`). ~Fixes were needed in two places, and~
~both times it was about uniquifying type parameters by prepending an~
~underscore (once when looking up a trait method through a class type,~
~and once when flattening the trait).~ **Edit:** Following @albertnetymk's
suggestion, I moved the alpha conversion (adding an underscore to the 
beginning of all parameters of trait methods) to the prechecker, meaning
that the environment and preprocessor are left unchanged. The underscores
are added to the parameters of *all* trait methods, and then removed when
showing the type. 

A test has been added.

Fixes #791.


albertnetymk:
On trait flattening, will universally prefixing all type vars simply the logic a bit? (After all, they are in different scope.) This way, `classMethodLookup` may be keep intact?

supercooldave:
Type variables can be alpha renamed at will.

EliasC:
> On trait flattening, will universally prefixing all type vars simply the logic a bit? (After all, they are in different scope.) This way, `classMethodLookup` may be keep intact?

Unconditionally prefixing type parameters is indeed possible and has been pushed. Thanks for the suggestion. Unfortunately, `classMethodLookup` must still be changed for the case when we do type inference:

```
trait T
  def id[t](x : t) : t
    x
  end
end

class C[t] : T
end

...

val c = new C[String]
c.id(42)
```

When doing inference for the final line, we have the binding `t -> String` (where `t` is the parameter of `C`), but inference will also find the binding `t -> int` (where `t` is the type parameter of the method), leading to a clash. The change in `classMethodLookup` will avoid the clash by renaming the `t` of `id` to `_t`.

albertnetymk:
I may have overlooked sth, but I thought after preprocessing (with renaming above), `C` becomes:

```
class C[t] : T
  def id[_t](x : _t) : _t
    x
  end
end
```
Therefore, `classMethodLookup` doesn't see any clashes.

EliasC:
Preprocessing is done before code generation but *after* typechecking, which is when `classMethodLookup` is used. We could possibly do the same kind of renaming before typechecking (in the desugarer), but that would affect error messages. Maybe that is something we could live with?

albertnetymk:
The alpha conversion in `Preprocessor` is using postfix instead of prefix; changing `Foo[t]` to `Foo[t_]` results into C error.

The logic introduced in `classMethodLookup` and `Preprocessor` is identical, so I think it's best to move it early in the pipeline. As for error msgs, maybe we can change `data Type` so that on `show t:Type`, it prints the original string representation instead of whatever used internally.

EliasC:
> The alpha conversion in `Preprocessor` is using postfix instead of prefix

Oops! Fixed now (and rebased). Thanks!

I'll see if it's possible to move the renaming to earlier in the pipeline.

EliasC:
Implemented the change suggested by @albertnetymk, and the code is a lot cleaner now! Thanks!

albertnetymk:
It looks good to me. Will squash-and-merge at 23:00 if no objections.

PR #801, opened 2017-05-05
Pull from ponyrt 54fbf3c

albertnetymk:
Two identified major change:

- `pony_alloc_final` use two args, and finalizer function is specified
  in `pony_type_t`

- `hash.c` uses robin hood hash algorithm


EliasC:
I've only looked at the changes to the Encore specific files, but since they look reasonable and the tests pass, I'm in favour of merging this. Will do so when someone else has also approved it.

kikofernandez:
I am a bit skeptical to merging this before the meeting in Brussels. This PR has only been tested once and I am not even sure if someone has run all the tests on a OS X. Before this gets merged, it should be discussed (tomorrow) and tested on OS X.

albertnetymk:
It would be good if this could be done before this month ends.

EliasC:
@kikofernandez Are you fine with merging this now that the review is over?

kikofernandez:
yes, you can merge this after 15.00 today

PR #803, opened 2017-05-11
Remove C-warning when using non unit-typed loops

### LINKED ISSUE #759: Gets C warning when for loop has non-unit type

EliasC:
This commit gets rid of a C-warning when the body of a loop does
not have the type `unit`. No test has been added, as warning
output cannot be tested.

Fixes #759.

### DIFF
--- a/src/back/CodeGen/Expr.hs
+++ b/src/back/CodeGen/Expr.hs
@@ -663,26 +663,17 @@ instance Translatable A.Expr (State Ctx.Context (CCode Lval, CCode Stat)) where
                          result <- Ctx.genNamedSym (fromJust sym)
                          return (Var result, Assign (Decl (translate retTy, Var result)))
 
-  translate w@(A.DoWhile {A.cond, A.body}) =
-      do (ncond,tcond) <- translate cond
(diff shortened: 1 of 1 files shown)


albertnetymk:
The commit msg doesn't reflect the actual content accurately; it changes the type of all `loop` constructs.

EliasC:
> The commit msg doesn't reflect the actual content accurately; it changes the type of all `loop` constructs.

What do you mean? All loops are already given the type `unit`. This commit just removes the generated code that assigns the result of the body to a variable. Since the result will always be interpreted as having the type `unit`, we can just return the unit value directly (as in this commit).

albertnetymk:
My mistake; I thought loops still get the type of the last expr. Thanks for the clarification.

albertnetymk:
Merge at 16:00 if no objections.

PR #804, opened 2017-05-11
Print file and line number for uncaught matches

### LINKED ISSUE #733: "no matching clause  was found" doesn't report encore program location
The following error message doesn't related to the actual encore source location.

```
*** Runtime error: No matching clause was found ***
```

It should.

EliasC:
This commit adds to the error message the position of a match
expression when no case matches. A test has been added.

Fixes #733.


albertnetymk:
Merge at 16:00 if no objections.

Issue #806, opened 2017-05-16
Closure's body does not create the future by async method call


PhucVH888:
Closure does not create the future that is created by asynchronous method call in its body.
```
active class Main
  def main() : unit
    val fres = (fun (x : int) : Fut[int] => this ! inc(x))
    println("{}", get(fres(42)))
  end

  def foo() : int
    10
  end

  def inc(x : int) : int
    x + 1
  end
end
```
The issue won't happen if the `await()` is used.
```
    ...
  def main() : unit
    val fres = (fun (x : int) : Fut[int] => this ! inc(x))
    val t = fres(42)
    await(t)
    println("{}", get(t))
  end
    ...
```

EliasC:
I don't understand what is the problem here. Looking at the generated code for the closure, the future seems to be created just fine:
```
static value_t _enc__closure_fun_closure0(pony_ctx_t** _ctx, pony_type_t** runtimeType, value_t _args[], void* _env)
{
  ENC_DTRACE1(CLOSURE_ENTRY, (uintptr_t)*_ctx);
  int64_t _enc__arg_x = (_args[0]).i;
  _enc__active__bug_Main_t* _enc__field_this = (*((struct _enc__env_closure0*) _env))._enc__field_this;
  check_receiver(_enc__field_this, " ! ", "this", "inc", "\"bug.enc\" (line 8, column 49)");
  pony_type_t* _tmp_0[] = {};
  future_t* _fut_1 = _enc__method__bug_Main_inc_future(_ctx, _enc__field_this, NULL, _enc__arg_x);
  int64_t _tmp_2 = future_get_actor(_ctx, _fut_1).i;
  ENC_DTRACE1(CLOSURE_EXIT, (uintptr_t)*_ctx);
  return ((encore_arg_t) {.i = _tmp_2});
}
```
(the call to `_enc__method__bug_Main_inc_future` is where the future is created). 

However, the program will deadlock as it has an actor blocking on a future which the actor itself is supposed to fulfill. When using `await`, you let the actor handle other messages before continuing, so the deadlock is avoided.

supercooldave:
The problem is deadlock, as you say. We worked that out a little later.

EliasC:
I don't think this is an Encore bug. If you cause a deadlock, your program will deadlock.

PhucVH888:
@EliasC I see. I thought it was a bug. The fix is no longer necessary, and discarded soon.

supercooldave:
Closing, as this was not a bug.

PR #808, opened 2017-05-17
Fixed using future in closure's body

PhucVH888:
This PR fixes the issue #806 .

Here is the issue: when a future is used in the closure's body, it may causes a deadlock.

### Example 1: 
```
active class Main
  def main() : unit
    val fres = (fun (x : int) : Fut[int] => this ! inc(x))
    println("{}", get(fres(42)))
  end
  def inc(x : int) : int
    x + 1
  end
end
```
### Example 2
```
active class Main
  def main() : unit
    val fres = fun (x : int) : int => get(this ! inc(x))
    println("{}", fres(42))
  end

  def inc(x : int) : int
    x + 1
  end
end
```
The issue won't happen if the await() is used.
``
    ...
  def main() : unit
    val fres = (fun (x : int) : Fut[int] => this ! inc(x))
    val t = fres(42)
    await(t)
    println("{}", get(t))
  end
    ...
```


kikofernandez:
As far as I see, this solution penalizes all calls to `get`, even the ones that would not cause deadlock. Is there any other way of dynamically checking if the call on `get` would cause deadlock and, if so, run this code that you generate?

EliasC:
If I understand correctly, this PR inserts an implicit `await` whenever `get` is used inside a closure, and when calling a closure which returns a future. I think this is bad for several reasons: 

* First of all, it interrupts the current behavior in a way that is invisible to the programmer. An `await` means that other messages may be handled by the current actor until the future is fulfilled. In the following code, I would expect the field `fld` to stay unchanged, but since there is now an implicit `await`, I don't have that guarantee:
    ```
    active class A
      var fld : int
      def m() : unit
        val old = this.fld
        val f = fun () => get((new Foo) ! foo())
        f()
        assertTrue(this.fld == old) -- May fail!
      end
    end
    ```
    Similarly, I lose the ability to call a closure that returns a future and pass that future to someone else while continuing my work. For example, in the following code, the current actor will wait for each future to be fulfilled before sending the next one, killing all parallelism:
    ```
    def distribute(getNext : () -> Fut[int]) : unit
      for worker <- this.workers do
        val fut = getNext()
        -- Implicit barrier, where we wait for the future to be fulfilled
        worker ! work(fut)
      end
    end
    ```
* Secondly, it seems a bit ad-hoc that I get one behavior when calling a closure, and another one when calling a global function. For example, in the following code, I will have different blocking behavior when calling `foo` and when calling `f`, even though they refer to the same function:
    ```
    fun foo() : Fut[int]
      ...
    end
    fun bar() : unit
      val f = foo
      foo() -- No interruption
      f() -- Implicit await
    end
    ```
* Lastly, I'm not sure I think this is a bug in the language. If you cause a deadlock in your program, that's a programmer error, not a language issue. If we want to prevent deadlocks, we should come up with a way to get rid of them completely, not just for very specific cases like this.

supercooldave:
I agree with @kikofernandez and @EliasC. This changes the semantics, so should be rejected.
If the problem is that you have a test case that doesn't test `forward` as you would like, why not add `await` in the test case?

PhucVH888:
I see. Thanks for all comments. I thought it was a bug, so the fix is no longer necessary, and closed now.

PR #811, opened 2017-05-31
cast print

kikofernandez:
Casting a debugging printout to make it work on Ubuntu.
I am surprised that the CI didn't caught this error. It happens for Ubuntu Trusty (32 bits version)

### DIFF
--- a/src/runtime/stream/stream.c
+++ b/src/runtime/stream/stream.c
@@ -55,7 +55,7 @@ __attribute__ ((unused))
 static void scons_print(struct scons *scons){
   printf("struct scons@%p{\n", scons);
   printf("  eos     = %s\n", scons->eos? "true": "false");
-  printf("  element = %ld\n", scons->element.i);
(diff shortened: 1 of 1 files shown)


albertnetymk:
I think the reason why CI didn't complain is that it's 64-bit OS. The portable way of printing `intptr_t` is shown at https://stackoverflow.com/questions/5795978/string-format-for-intptr-t-and-uintptr-t

kikofernandez:
ok, solved. Tested on Mac OS X and Ubuntu

albertnetymk:
Wouldn't `PRIxPTR` produce hex format? The original (when it works) produces decimal format. Is this change intended?

kikofernandez:
I completely misunderstood this website http://www.cplusplus.com/reference/cinttypes/ 
It has been fixed to be PRIdPTR

albertnetymk:
Squash-and-merge at 22:00 if no objections.

PR #761, opened 2017-03-29
Added the BankTransaction Savina benchmark

EliasC:
This will not work on Mac OSX until #760 is merged!

PR #812, opened 2017-06-01
Added the BankTransaction Savina benchmark

EliasC:
This PR subsumes #761.


supercooldave:
It would definitely be interesting to see if the run-time will improve with the new future implementation. And/or whether this is a better way of doing this -- there are a lot of gets at the end of methods where the value of returned by the future is not used. Perhaps there's a better way.

supercooldave:
(My comment shouldn't affect this PR.)

EliasC:
> there are a lot of gets at the end of methods where the value of returned by the future is not used

The way this is done in the original benchmark is through selective message receives to make sure that no other transactions are started before the current one finishes. Benchmarking against new futures or the implementation of the `atomic` block would indeed be interesting! (ping @duckyduckmaster @M34ris).

modules/standard/Random.enc
```diff
@@ -37,4 +37,9 @@ read class Random
 
     END
   end
```
supercooldave, line 41: Does your return key not work?

supercooldave (commented):
One terribly minor comment, but there should be newlines between function definitions.

supercooldave:
One minor comment. I'm happy to accept this into the fold after it is addressed.

supercooldave:
Merging at 16:00 unless there are objections.

modules/standard/Random.enc
EliasC: This code was translated automatically at some point, and the pretty-printer adds an extra newline before an `END` (for some reason)

EliasC: (and does not add newlines between methods)

supercooldave: I fixed it all up by hand. 
(Didn't the return key work when writing the pretty printer? :smiley:)

EliasC: Thanks! 😃

PR #813, opened 2017-06-01
Fix bug when printing non-sugared strings

EliasC:
Printing a string created using `new` would break:

```
println(new String(EMBED (EMBED char* END) "foo"; END))
```

This commit fixes this, adds a test for it, and also improves the
error message of `Meta.hs: No type given to node` so that it
prints the source position of the node that was not given a type,
which should help when debugging similar errors in the future.


kikofernandez (approved):
The changes look sensible and showing is position for future references sounds good as well. Merging at 12.00 today

PR #821, opened 2017-06-13
new conc future api

duckyduckmaster:
An implementation of multiple classes of lock-free futures, in order to fix the scalability problems of using only one class with mutex-locks. There are three different future classes available which are defined in three different source files: 
* future.c : futures with multiple get, await and chain operations.
* vanilla_future.c : futures with only one get operation.
* poly_vanilla_future.c : futures with multiple get operations.

Currently the API of the future.c file is the only used and further code is needed in order to enable vanilla and poly vanilla futures, such as compiler analysis: 
I forgot to include tests so they may be put inside another PR.


kikofernandez:
could you provide some brief explanation about the different kinds of futures?

src/runtime/future/vanilla_future.c
```diff
@@ -0,0 +1,168 @@
...
+
+static inline void vanilla_future_discharge(pony_ctx_t **ctx, vanilla_future_t *fut) {
+  if (__sync_bool_compare_and_swap(&(fut->blocking), true, false)) { 
+     perr("Unblocking");
+     actor_set_resume((encore_actor_t *)fut->actor);
+     pony_schedule(*ctx, fut->actor);
+  }   
+}
+
+void handle_future(encore_actor_t *actor, pony_ctx_t *futctx, void * pony_node)
```
kikofernandez, line 134: this function seems to be used for all futures. Is this correct?
If so, shouldn't it be placed in a more general future file, instead of in the vanilla future?
(I may be missing the point where this is called and if this is always called)

duckyduckmaster: Yes, it is correct that this function is used for all futures, the reason why its placed inside the vanilla future file is because the vanilla future is the future which is most often used, and therefore most calls to handle_future can be inlined.

src/runtime/future/poly_vanilla_future.c
```diff
@@ -0,0 +1,141 @@
...
+    ENC_DTRACE2(FUTURE_UNBLOCK, (uintptr_t) *ctx, (uintptr_t) fut);
+  }
+  ENC_DTRACE2(FUTURE_GET, (uintptr_t) *ctx, (uintptr_t) fut);
+  return fut->value;
+}
+
+static inline void poly_vanilla_future_block_actor(pony_ctx_t **ctx, poly_vanilla_future_t *fut)
+{
+  perr("future_block_actor");
+  if (__atomic_load_n(&(fut->fulfilled), __ATOMIC_SEQ_CST)) {
```
kikofernandez, line 104: Is this condition needed at all? The function above (`poly_vanilla_future_mk`) seems to do this checking.

kikofernandez, line 104: Does the `__atomic_load_n` work similar to acquire and release semantics, i.e., the state between acquire and release can be thought as a "transaction" visible to other threads?

src/runtime/future/future.h
```diff
@@ -79,4 +121,37 @@ void future_register_callback(pony_ctx_t **ctx,
  * puts on hold the processing of this message.
  */
 void future_await(pony_ctx_t **ctx, future_t *fut);
+
+extern pony_type_t future_type;
+
+void future_discharge(pony_ctx_t **ctx, future_t *fut);
+
+/* VANILLA FUTURE API */
```
kikofernandez, line 129: It would be great if the definition of *vanilla* futures et al is specified in the header files (or somewhere).
From the PR description:
>future.c : futures with multiple get, await and chain operations.
vanilla_future.c : futures with only one get operation.
poly_vanilla_future.c : futures with multiple get operations.

it's not clear to me if a vanilla future has a single get operation or if it means that I cannot get multiple times from it (which could happen if there is an optimisation to reuse the future, e.g.)

src/runtime/future/vanilla_future.c
```diff
@@ -0,0 +1,168 @@
...
+    ENC_DTRACE2(FUTURE_UNBLOCK, (uintptr_t) *ctx, (uintptr_t) fut);
+  }
+  ENC_DTRACE2(FUTURE_GET, (uintptr_t) *ctx, (uintptr_t) fut);
+  return fut->value;
+}
+
+static inline void vanilla_future_block_actor(pony_ctx_t **ctx, vanilla_future_t *fut)
+{
+  perr("future_block_actor");
+  if (__atomic_load_n(&(fut->fulfilled), __ATOMIC_SEQ_CST)) {
```
kikofernandez, line 103: same as in poly_vanilla. Is this if needed?

kikofernandez, line 103: many of these methods seem to be exactly the same.
Would it make sense to have a macro that defines these common methods and the specialised methods in the files?

src/runtime/future/future.c
```diff
@@ -140,57 +89,38 @@ void future_trace(pony_ctx_t *ctx, void* p)
...
 // ===============================================================
 // Create, inspect and fulfil
 // ===============================================================
 
-//
-// a future has a lock attached to it. actors may acquire the same lock
-// multiple times (re-entrant locks). this does not happen when we work with plain Encore
-// however, re-entrant locks are necessary for the ParT runtime library.
-//
-// Use case:
```
kikofernandez, line 162: not sure if the removal of this may affect ParTs

src/runtime/future/future.c
```diff
@@ -360,107 +240,140 @@ void future_register_callback(pony_ctx_t **ctx,
...
+  ENC_DTRACE2(FUTURE_FULFIL_START, (uintptr_t) *ctx, (uintptr_t) fut);
+  fut->value = value;
+  __atomic_store_n(&(fut->fulfilled), true, __ATOMIC_SEQ_CST);
+  future_gc_send_value(*ctx, fut);
+  future_discharge(ctx, fut);
+  ENC_DTRACE2(FUTURE_FULFIL_END, (uintptr_t) *ctx, (uintptr_t) fut);
+}
 
-void future_await(pony_ctx_t **ctx, future_t *fut)
+void future_discharge(pony_ctx_t **ctx, future_t *fut)
```
kikofernandez, line 307: I am not sure of what the naming means here. Maybe it's best if you provide some more information.
At the same time, I get the feeling that this *discharge* is possibly doing too many things. As I understood:
1. Goes through the list of blocked actors and schedules them
2. Starts running chained computations attached to the future, following a LIFO order. I would have expected that computations that are attached first run before the ones attached later (if the chaining operation happens in the same actor, and it's fair to assume any order if the future is shared among actors that may chain computations in this same future)

kikofernandez (changes requested):
Before this can be merged, I would like for the author to reply to the feedback :)

duckyduckmaster:
1. The condition isn't needed
2. I am not that into how different atomic operations work, I just heard
somewhere that using the flag  __ATOMIC_SEQ_CST is the most strict option,
reducing the potential errors encountered.
3. Many methods behave similarily, like in this case. I think I sacrificed
alot of simplicity in return for an obscure gain in speed from micro-opts,
something which could be questioned in the long run.  I think both macros
and branch structures could be helpful.
4. The comment could stay for educational purposes I think, as long as it's
pointed out that locks are no longer used.
5. Yes, I agree. for fairness the stack with closures should be a queue
with a FIFO pattern. The same principle could be applied to the stack with
actors.

2017-09-15 13:36 GMT+02:00 Kiko <notifications@github.com>:

PR #822, opened 2017-06-14
Core Libs

supercooldave:
This PR focusses on Encore libraries.
- It includes Lucas's libraries, converted to the latest syntax, bug-fixed, refactored.
- Some additional libraries, in particular dealing with Arrays and Maybe type
- A small Math library
- A StringBuilder
- Char
- Some minor contributions to other libraries.
- Some refactoring of existing library structure
- Consequent fixes to test suite and other code.
- Added HashCode to String
- remove equals from String (was deprecated)
- minor modification to pattern matching due to removal of String.equals()
- updated to use bounded polymorphism

Please comment on organisation of libraries, names chosen with each module, etc.

Note that the test suite for some libraries was automatically translated from the old syntax, and they look pretty horrendous.


supercooldave:
Seems to have a bug related to the new return type rule.
I'll track it down and swat it.

PR #827, opened 2017-06-21
Modified type of return for better programmer experience

### LINKED ISSUE #824: [RFC] Type of `return` is wrong
The type rule (and corresponding implementation) of the type of `return` is as follows:
```
     E |- expr : t
     E |- currentMethod : _ -> t
    -----------------------------
      E |- return expr : t
```
I argue that this is wrong, and the type should be `unit` or, even better, anything.
```
     E |- expr : t
     E |- currentMethod : _ -> t
    -----------------------------
      E |- return expr : t'
```

Consider the following example (out of context):
```
if foo then
    return 10
else
    this.skip()
end
```
where `this.skip()` has type `unit`.

This does not type check, even though the only one of the branches actually returns a result.

I would argue that making the type anything (which is probably `Bottom` in the current implementation) rather than `unit` will allow for more flexibility, permitting the following example to compile:
```
  def bar() : String
    "bar"
  end

  def foo() : Maybe[int]
    if true then
      return Nothing : Maybe[int]
    else
      this.bar()
    end
    Just(11)
  end
```

The fix should be easy, but I'm not sure what the consequence will be. My understanding is that Bottom type does not play nicely.

Similar comments apply to `forward`.

supercooldave:
This commit changes the type rule for `return` from
```
     E |- expr : t
     E |- currentMethod : _ -> t
    -----------------------------
      E |- return expr : t
```
to
```
     E |- expr : t
     E |- currentMethod : _ -> t
    -----------------------------
      E |- return expr : bottomType
```
The reason is that bottom type  matches any type (in principle). In addition, the PR also loosens the rules for dealing with `bottomType` to make it more useful, without any negative consequences (as far as the test and stress suites are concern).

Because `return` does no result in a value in the current expression, its type in theory can be anything. This is useful to type `if` statements that use return in one branch and do something (imperative) in the other branch:
```
if foo then
    return 10
else
    this.skip()
end
```
One no longer needs to add in a fake value to trick the type system into accepting that the second branch has type `int`.

This fix broke one of the tests (`encore/return/return.enc`). The relevant part is:
```
  def c() : unit
    let
      x = return(())
    in
      ()
    end
  end
```
With the fix, the compiler cannot work out the type of `x`. If such code needs to be written, then adding a type annotation to x (`x : unit`) is sufficient to make it compile.

This is better, in my opinion, than adding unused values in branches of if statements to make them compile.

Added two additional tests to deal with if-statements/match clauses with returning and not returning bodies.

Fixes #824.


supercooldave:
The same idea could/should apply to `forward` (@PhucVH888).

supercooldave:
I noticed that it broke some of the stress tests.
I'll see what the story is and fix.

supercooldave:
The problem seems to be methods/functions that do not produce a result in the normal manner, such as 
```
def broke() : int
  if true then
    return 10
  else
    return 11
  end
end
```
Though the following works perfectly:
```
def broke() : int
  return 10
end
```

supercooldave:
Problem mentioned above is now fixed.

src/types/Typechecker/TypeError.hs
```diff
@@ -510,7 +510,8 @@ instance Show Error where
         printf "Unbound function variable '%s'" (show name)
     show (NonFunctionTypeError ty) =
         printf "Cannot use value of type '%s' as a function" (show ty)
-    show BottomTypeInferenceError = "Cannot infer type of 'Nothing'"
+    show BottomTypeInferenceError = "Not enough information to infer the type.\n" ++ 
+        "Try adding more type information."
```
EliasC, line 514: This is great! It has been annoying me a long time.

EliasC (approved):
I think this looks good! I pushed a tiny addition to improve the error message when matching on `Nothing` (requiring using type annotations). If @kikofernandez approves, this can be merged this afternoon.

EliasC:
Ping @kikofernandez. If you check my tiny addition, I can vouch for the rest :)

kikofernandez:
your addition looks good. You can merge

kikofernandez (approved):
This provides a better programming experience. The PR is ready for merging

PR #828, opened 2017-06-22
Maybe, Par etc are no longer reserved

supercooldave:
This PR makes Maybe, Fut, Par and Stream no longer reserved names in the parser. This means that the names Maybe etc can be used more freely within the code, in particular (and most importantly), as the name of a module. (They cannot be used as the name of a new type.) This will be useful when creating a module of functions for Maybe, which can now be called `module Maybe`.

Additional checking has been added to ensure that new classes etc are not called Maybe etc.

In addition, StreamLib was renamed to Stream (and reorganised slightly).

### DIFF
--- a/modules/standard/Stream/Chain.enc
+++ b/modules/standard/Stream/Chain.enc
@@ -1,7 +1,7 @@
--- This file was automatically converted by encorec
+module Chain
 
-module StreamChain
 typedef Scons = EMBED struct scons* END
+
 fun chain[a, b](sa : Stream[a], f : a -> b) : Stream[b]
   let
     futa = EMBED (Fut[Scons])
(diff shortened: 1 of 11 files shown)


kikofernandez:
If I can create a new type `Fut` or `Par`, how do you handle primitive operations in those types, e.g. `~~>` or `>>`? How do you know that the new `Fut` type has the same semantics of futures as in Encore? Same for ParT

supercooldave:
@kikofernandez You cannot make a new `Fut` or `Par` type. The names `Fut` and `Par` etc are no longer keywords (reserved) in the parser, so they can be used in other parts of the code.

The reason for the change is so that the stream lib can be called `module Stream` instead of `module StreamLib` and the Maybe module can be called `module Maybe` instead of `module Maybee`.

kikofernandez:
Sorry, I saw the `MaybeRedefine.enc` example and I thought that you allowed that possibility. All clear now :)

kikofernandez:
I can review this on Monday unless someone else does before me

kikofernandez (approved):
The changes are sensible. I will merge today at 15.00

PR #829, opened 2017-06-22
Cleaned up stress tests.

supercooldave:
Added entries into IGNORED_FILES.grep to reduce number of warning produced when running stress tests.

Made one Makefile a little more robust.

### DIFF
--- a/src/tests/stress/savina/11.BndBuffer/Makefile
+++ b/src/tests/stress/savina/11.BndBuffer/Makefile
@@ -1,8 +1,10 @@
 
 main:
-	encorec -O3 main.enc -I ../../modules/standard
+	../../../../../release/encorec -O3 main.enc
 
 all: main
 
 run: main
 	/usr/bin/time ./main
+
+clean:
--- a/src/tests/IGNORED_FILES.grep
+++ b/src/tests/IGNORED_FILES.grep
@@ -1,5 +1,5 @@
 # This file support line comments.
-# Note that the # must appear first on the line
+# Note that the comment character # must appear in the first position on a line
 
 #
 # Modules used by 'modules' tests
--- a/src/tests/IGNORED_FILES.grep
+++ b/src/tests/IGNORED_FILES.grep
@@ -1,13 +1,6 @@
 # This file support line comments.
 # Note that the # must appear first on the line
 
-#Pedsim util modules should not be considered tests
-stress/pedsim/Ped_util/Agent_passive
-stress/pedsim/Ped_util/Global_funs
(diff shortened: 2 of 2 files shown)


albertnetymk:
Listing each directory explicitly under `stress` seems an unnecessary burden to maintain, for we *know* that they are stress tests, not ordinary tests. (Doesn't the ignore scheme in the testing script support whole directory exclusion?)

Using explicit relative path while invoking the compiler seems a bit odd as well. `../../../../../release/encorec -O3 main.enc`

supercooldave:
If you don't mention the directories, then the screen is filled with warnings, though the tests will still pass.
It was decided not to suppress such warnings and to force you, the Encore developer, to explicitly add entries to IGNORED_FILES.grep, to avoid accidentally submitting a test file (.enc) without the corresponding other files.

I added the explicit path to the encore compiler to make sure that it was running the right one — perhaps this is redundant and handled already by the testing script.

src/tests/stress/savina/11.BndBuffer/Makefile
```diff
@@ -1,8 +1,10 @@
 
 main:
-	encorec -O3 main.enc -I ../../modules/standard
+	../../../../../release/encorec -O3 main.enc
 
 all: main
 
 run: main
 	/usr/bin/time ./main
+
```
kikofernandez, line 9: Is this `Makefile` needed by the CI?

kikofernandez (commented):
Looks good but I won't merge until the author confirms if a `Makefile` is needed to do the stress testing. If the author fixes this issue before 15.00, I'll merge today

src/tests/stress/savina/11.BndBuffer/Makefile
supercooldave: Yes, because there is no other way to pass the -O3 flag to the compiler.
(Whether the -O3 flag is really required is a question for the BoundedBuffer implementor.)
The reason for tidying up the Makefile was to reduce errors in the output, and increase faith in `make stress`.

supercooldave:
@kikofernandez Ping!

supercooldave:
@kikofernandez Are you fishing for more cake?

kikofernandez:
CI fails

_Sent from my Samsung GT-I9506 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

supercooldave:
CI fails because I keep updating src/tests/IGNORED_FILES.grep whenever I add new (stress) tests.
I'll fix soon and checkout that Makefile.

Issue #831, opened 2017-06-27
Access to arrays is broken for non-array accesses


kikofernandez:
The problem is that the parser considers an `ArrayAccess` anything that matches `x(y) = z`. The typechecker goes along and considers that the AST node given by the parser is an array, and tries to typecheck it incorrectly, crashing the compiler.

This has been shown in #823 and also in this example:

```ruby
active class Main
  def main(): unit
    val x = 3
    x(1) = 2
  end
end
```

Error thrown:

```
encorec: No match in record selector argTypes
```

PR #832, opened 2017-06-28
fix array access

### LINKED ISSUE #823: Array access operator applied to string results in internal compiler error
Compiling the following class
```
active class Main
  def main() : unit
    val str = "foo"
    str(0)
  end 
end
```
results in the internal error
> encorec: No match in record selector argTypes

I suspect this is due to trying to use array access `(0)` on the string.
(It would be nice to use `str(n)` to access elements of the string.)

kikofernandez:
This commit fixes accesses to an array when the type is different from an array; closes #823 and #831 

This test case used to crash the compiler
```
active class Main
  def main(): unit
    val i = 4
    i(0) = 3
  end
end
```

With this commit, the compiler informs you that it expects an array of a function call and you gave it an int type.

### DIFF
--- a/src/tests/encore/basic/arrayAccess.fail
+++ b/src/tests/encore/basic/arrayAccess.fail
@@ -1,1 +1,1 @@
-Expected an array access or a function call but found expression of type 'int'
+Expected an array or a function call but found expression of type 'int'
--- a/src/tests/encore/basic/arrayAccess.fail
+++ b/src/tests/encore/basic/arrayAccess.fail
@@ -1,0 +1,1 @@
+Expected an array access or a function call but found expression of type 'int'
--- a/src/types/Typechecker/Typechecker.hs
+++ b/src/types/Typechecker/Typechecker.hs
@@ -824,8 +824,6 @@ instance Checkable Expr where
                   Just (qname2, ty2) -> return defName
                   Nothing -> tcError $ WrongNumberOfFunctionArgumentsError
                               qname (length argTypes) (length args)
-          unless (isArrowType ty) $
-            tcError $ NonFunctionTypeError ty
           (eArgs, returnType, typeArgs) <-
             if null typeArguments
             then inferenceCall fcall typeParams (take actualLength argTypes) resultType
--- a/src/types/Typechecker/Typechecker.hs
+++ b/src/types/Typechecker/Typechecker.hs
@@ -807,39 +807,40 @@ instance Checkable Expr where
           doTypecheck ArrayAccess{emeta
                                  ,target = VarAccess{emeta, qname}
                                  ,index = head args}
-      else do
-        let typeParams  = getTypeParameters ty
-            argTypes    = getArgTypes ty
-            resultType  = getResultType ty
-            actualLength = length args
-            expectedLength = length argTypes
-            defName = qname'{qnlocal = Name $ "_" ++ show qname ++ show (expectedLength - actualLength)}
-
-        calledName <-
-          if (actualLength == expectedLength)
-            then return qname'
+      else
+        if (isArrowType ty) then do
+          let typeParams  = getTypeParameters ty
+              argTypes    = getArgTypes ty
+              resultType  = getResultType ty
+              actualLength = length args
+              expectedLength = length argTypes
+              defName = qname'{qnlocal = Name $ "_" ++ show qname ++ show (expectedLength - actualLength)}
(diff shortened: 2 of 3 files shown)


src/types/Typechecker/Typechecker.hs
```diff
@@ -807,39 +807,40 @@ instance Checkable Expr where
...
+                     tcError $ WrongNumberOfFunctionTypeArgumentsError qname
+                               (length typeParams) (length typeArguments)
+              typecheckCall fcall typeParams (take actualLength argTypes) resultType
+          return $ setArrowType ty $
+                   setType returnType fcall {args = eArgs,
+                                             qname = calledName,
+                                             typeArguments = typeArgs}
+        else do
+          tcError $
+             ExpectingOtherTypeError "an array access or a function call" ty
```
EliasC: Minor: Should this be `"an array or a function"`? This way the error message becomes
> Expected an array or a function but found expression of type 'int'

reflecting that the type error is not the whole expression, but only the sub-expression that is being called.

EliasC (approved):
Looks straightforward enough (especially when reviewed with `?w=1`). I only have a minor comment on the error message, but I would be fine with merging it anyway. Will merge this afternoon unless anyone objects.

kikofernandez:
thanks! I have remove the `access` word from the error message

EliasC:
Have you also pushed?

src/tests/encore/basic/arrayAccess.fail
```diff
@@ -0,0 +1 @@
+Expected an array access or a function call but found expression of type 'int'
```
EliasC: This test needs to be updated as well

albertnetymk:
It seems that `unless (isArrowType ty)` is guarded by an `if` with the same condition.

src/types/Typechecker/Typechecker.hs
```diff
@@ -807,39 +807,40 @@ instance Checkable Expr where
...
+            if (actualLength == expectedLength)
+              then return qname'
+              else do
+                result2 <- findVar defName
+                case result2 of
+                  Just (qname2, ty2) -> return defName
+                  Nothing -> tcError $ WrongNumberOfFunctionArgumentsError
+                              qname (length argTypes) (length args)
+          unless (isArrowType ty) $
+            tcError $ NonFunctionTypeError ty
```
EliasC: Here is the thing that could go according to @albertnetymk. I agree.

kikofernandez: good catch @albertnetymk and @EliasC. Fixed!

albertnetymk:
OK, then it makes sense to restructure the flow to be sth like the following to "return early" for abnormal case, IMO.

```haskell
   unless (isArrowType ty) $
    error
   -- normal case follows
```

kikofernandez:
@albertnetymk I thought about it and considered that the majority of the time the programmer knows that an array is an array, so I was favouring the other pattern, where the developer is correct in most occasions. I have no problems going for your suggestion though.

(@albertnetymk it would be great if your comments are actually part of the review or marked in the appropriate lines, as the ones from @EliasC)

EliasC:
I agree with Kiko. Having the failing case last makes sense to me.

albertnetymk:
By "return early", a level of indentation is removed, which is considered a decrease in certain complexity metrics. Surely, this is very subjective; the author should make the judgement call.

EliasC:
> By "return early", a level of indentation is removed

The same "complexity reduction" could be achieved by writing

```
else if ... then
  ...
```
instead of
```
else
  if ... then
    ...
```

albertnetymk:
> The same "complexity reduction" could be achieved by writing

I don't think that's true in Haskell, because `body` needs to be indented further than `if`.

kikofernandez:
This is a matter of taste and style. We all have provided good reasons for choosing one option or the other. Can we please move past the `if-else`?

EliasC:
I took an executive decision and merged this =)

albertnetymk:
The "early return" is about a taste of preference, but the indentation discussion is not.

My previous statement, "`body` needs to be indented further than `if`", is incorrect. The following is perfectly valid Haskell code:

```
f =
  if True then
 1
  else if True then
 2
  else
 3
```

kikofernandez:
Perfect valid Haskell code ≠ accepted code in Encore compiler

PR #834, opened 2017-06-29
Additional Savina Benchmarks

supercooldave:
This PR groups together Savina benchmarks that were previously submitted, but worked only for pre-Kappa. They have been fixed and now produce minimal output so they can be used in the stress test directory.

Credit should be given to Sophia, Einar and Tobias, who wrote these (but they also let them rot).

### DIFF
--- a/modules/standard/String.enc
+++ b/modules/standard/String.enc
@@ -314,6 +314,10 @@ read class String : Id + Eq[String](compare(),length,data)
       END
     end
   end
+  
+  def starts_with(s : String) : bool
+    this.find_from(s, 0) == 0
+  end
 
   -- Terrible code
   def delete(s:String) : String
--- a/src/tests/IGNORED_FILES.grep
+++ b/src/tests/IGNORED_FILES.grep
@@ -58,4 +58,13 @@ encore/stream/StreamIO
 # Module used by 'stress' tests
 # Should not be considered as a test
 #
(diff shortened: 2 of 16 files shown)


supercooldave:
Interesting. The recent array fix has resulted in an assertion failure in NQueens.
This could point to the bug Einar had trouble tracking down.

albertnetymk:
The failure was there before the merge of `array-fix`, FYI.

supercooldave:
@albertnetymk I knew the fix was related to arrays and the bug was related to arrays, so I put 2 and apple together and got yellow.

After looking in detail, the take away message is to every now and again run the compiler in debug mode to find more errors.

kikofernandez:
what's the status on this? I see a commit that seems to fix the error. Let me know when this is fixed and I will review and merge this.

supercooldave:
The bug in NQueens has been fixed.
NQueens still has the problem that it leaks memory because the bulk of the computation occurs within a single recursive method call. Running it with input 17 almost killed my machine. I'm not sure whether it is a high priority to fix the problem. I tried to using `forward`, but that revealed bugs in `forward`.

supercooldave:
In summary, just leave this open for a little longer. I'll see if I can find time to fix.

kikofernandez (approved):
this looks good. Merging today at 15.00

PR #838, opened 2017-06-30
Bounded polymorphism

### LINKED ISSUE #410: Need some way of restricting type parameters
As it currently stands, a polymorphically typed field can always be assigned _any_ type through the type parameter. This is a major problem for me, as I need to be able to declare some polytype fields invalid, but can't assign them the value `null` or bitwise or it with `1`, because it might not be a reference type.

This is also major problem when combining polymorphic types with the upcoming speculative types Elias is working on.

EliasC:
This PR adds support for bounded polymorphism. Whenever a formal type parameter is declared it may supply a trait type that this type must implement when instantiated:

```
fun equals[t : Eq[t]](x : t, y :t) : bool
  x.eq(y)
end
```

In this example, the function `equals` must be called with arguments which implement the `Eq` trait. Bounds may refer to other type parameters (as above). The same functionality applies to generic classes, traits and type synonyms, as well as polymorphic methods.

A substantial change to the typechecker is the separation of `resolveType` into two functions `resolveType` and `checkType`. `resolveType` will resolve a type and supply it with the correct namespace etc. `checkType` does the same but additionally makes sure that a type is well-formed. For example, if a class is defined as `class Foo[t : T]` and `Bar` is a class that does *not* implement `T`, the type `Foo[Bar]` can be resolved, but not checked. `resolveType` is used in the prechecker when we don't yet have enough information to check if types are well-formed. After the prechecker, only `checkType` should be used.

In the types module, the meta info of a type variable has been extracted to its own datatype so that equality can be defined to only be concerned with the names of the variables. Additionally there are commits to fix two bugs found along the way. Therefore the history should preferably not be squashed.

* A linear class with a non-manifest mode (e.g. `class C : linear T1 + read T2`) could be passed to a non-linear context (see test case `linear/nonManifest.enc`).
* A class that includes `Id` could not use `==` on `this` in an overridden method since the trait of the original method would not include `Id` (see test case `traits/id.enc`).

200 or so lines of this PR are test cases.

Fixes #410.


supercooldave:
I have been working with the PR for a little while already one a decent sized code base (Core Libs + JSON parser). I found a few bugs, but am now working bug free.

supercooldave:
@EliasC I'm going to look at this this morning, but it has a conflict now. Fix and ping me.

supercooldave:
This looks very good.
I've been experimenting with it while it was developed and a few bug fixes were resolved during that effort. 
Merging at 15:00.

PR #839, opened 2017-06-30
Fixed bug where polymorphic methods could not be overridden

EliasC:
This commit fixes a bug where polymorphic methods could not be
overridden if the names of the type parameters were
different (which they would always be since traits mangle their
names to avoid clashes).

### DIFF
--- a/src/types/Typechecker/TypeError.hs
+++ b/src/types/Typechecker/TypeError.hs
@@ -204,7 +204,7 @@ data Error =
   | CovarianceViolationError FieldDecl Type Type
   | RequiredFieldMismatchError FieldDecl Type Type Bool
   | NonDisjointConjunctionError Type Type FieldDecl
-  | OverriddenMethodTypeError Name Type Type
+  | OverriddenMethodTypeError Name Type Type Type
   | OverriddenMethodError Name Type Error
   | IncludedMethodConflictError Name Type Type
   | MissingMethodRequirementError FunctionHeader Type
@@ -391,10 +391,11 @@ instance Show Error where
         printf
           "Conjunctive traits '%s' and '%s' cannot share mutable field '%s'"
            (show left) (show right) (show field)
-    show (OverriddenMethodTypeError name expected trait) =
+    show (OverriddenMethodTypeError name expected trait actual) =
         printf ("Overridden method '%s' does not " ++
-                "have the expected type '%s' required by %s")
-               (show name) (show expected) (refTypeName trait)
+                "have the expected type '%s' required by %s.\n" ++
+                "Actual type is '%s'")
+               (show name) (show expected) (refTypeName trait) (show actual)
     show (OverriddenMethodError name trait err) =
         case err of
           FieldNotFoundError f _ ->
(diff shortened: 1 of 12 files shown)


src/types/Typechecker/Util.hs
```diff
@@ -285,13 +285,17 @@ subtypeOf ty1 ty2
     | isStackboundType ty1 =
         liftM (isStackboundType ty2 &&) $ unbox ty1 `subtypeOf` unbox ty2
     | isArrowType ty1 && isArrowType ty2 = do
-        let argTys1 = getArgTypes ty1
-            argTys2 = getArgTypes ty2
+        let typeParams1 = getTypeParameters ty1
```
kikofernandez, line 288: this seems correct but I would suggest to rename `ty1` and `ty2` to `actual` and `expected` (or similar). I think it would improve the readability and ease the understanding, since I (or anyone) would need to figure out which type is what.

src/types/Typechecker/TypeError.hs
```diff
@@ -391,10 +391,11 @@ instance Show Error where
         printf
           "Conjunctive traits '%s' and '%s' cannot share mutable field '%s'"
            (show left) (show right) (show field)
-    show (OverriddenMethodTypeError name expected trait) =
+    show (OverriddenMethodTypeError name expected trait actual) =
         printf ("Overridden method '%s' does not " ++
-                "have the expected type '%s' required by %s")
-               (show name) (show expected) (refTypeName trait)
+                "have the expected type '%s' required by %s.\n" ++
```
kikofernandez, line 396: I am guessing that we have tests for this error and we use a regex to match only part of it, is this right? If not, a test for this failing case should be written

kikofernandez (approved):
I have only two minor comments. One of them I think should be addressed (renaming name of variables) and the other one will be confirmed by @EliasC. As soon as I get a reply, I can merge this fix

src/types/Typechecker/Util.hs
supercooldave: I agree. I ran into same problem yesterday.

EliasC: I agree (`sub` and `super` are good names I think), but suggest that I do this in a different refactoring PR so that anyone looking at the history will find a clean change in this PR. OK?

src/types/Typechecker/TypeError.hs
EliasC: It seems we do not have a test for this! This PR does not introduce this error message (it only improves it), but I will add a test or two here anyway.

EliasC:
Pushed tests and a change to printing of function types to distinguish between `[t](int -> int)` and `int -> int`.

src/types/Typechecker/Util.hs
kikofernandez: Good. Don't take too long to create this refactoring :)

kikofernandez:
On another note, I was curious about this commit message:

>  The type of a function declared as `fun foo[a, b](x : a, y : b) : unit`
will now be printed as `[a, b]((a, b) -> unit)`. Such a higher
kinded type should currently only appear when talking about the
expected type of overridden methods. 

Why is it a higher kinded type function? I thought that higher kinded types abstract over types, such that you can use the untyped lambda calculus on top of types, e.g. do function application at the type level. I don't think our type system can do that, can it?

EliasC:
@kikofernandez The type `[a, b]((a, b) -> unit)` can be seen as a type-level function which takes two types `a` and `b` and produces the type `(a, b) -> unit`. It is true that Encore does not support higher-kinded types in general, but this is an example of where they show up under the hood.

PR #841, opened 2017-07-02
Fixed printing bottom problem

### LINKED ISSUE #836: printing Nothing crashes the compiler
Printing `Nothing` crashes the compiler. I believe this used to **not** be the case before.

Example:

```ruby
active class Main
  def main(): unit
    println(Nothing)
  end
end
```

Output:

```
encorec test.enc --run
encorec: I don't know how to translate Bottom to pony.c
```

supercooldave:
This PR converts BottomType to (void *) in code generation, rather than generating an error. My argument that this is safe is that anything of BottomType (or involving BottomType) never produces a value of BottomType, so the actual type could be anything. 

This fix was also tested with a larger program that had the same error in a different branch, but I was unable to reproduce that error in this branch.
Fixes #836.


EliasC (approved):
Straightforward and reasonable! Merging tomorrow.

PR #843, opened 2017-07-03
Make array subtyping invariant

### LINKED ISSUE #842: Array subtyping doesn't sound well.
If `B <: A`, the current encore compiler permits `[B] <: [A]`, which is unsound, as the following example demonstrates.
```
local trait A
end

local class B : A
  var f : int = 10
end

local class C : A
  var s : String = "Bar"
end

active class Main  
  def main() : unit
    val bb : [B] = [new B]
    val aa : [A] = bb -- this should not be allowed
    val c = new C
    aa(0) = c
    bb(0).f = 11  -- c.s will have been overwritten by 11
    println(c.s)  -- segmentation fault!
  end
end
```

EliasC:
This commit makes array subtyping invariant, as it should be (and not covariant, which is unsound). A test has been added.

Fixes #842.


kikofernandez:
Looks straight forward to me. Merging in!

_Sent from my Samsung GT-I9506 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #846, opened 2017-07-04
Improved type inference

### LINKED ISSUE #792: Restrictive forward's type
The `forward`'s type is too restrictive. When using nested `forward` in the following example, an extra variable `arg1` is necessary for passing the typecheck. A better way is to relax the `forward`'s type so that no extra variable is used to do the trick.
```
active class Base
  def base() : int
    42
  end
end

active class Bar
  def bar() : int
    24
  end
end

active class Foo
  def foo(b : bool) : int
    val arg1 = (new Bar) ! bar()
    val arg2 = (new Base) ! base()
    forward(if b then
              forward(arg1)
              arg1 -- required to make typecheck
            else
              arg2
            end)
  end
end

active class Main
  def main() : unit
    println("{}", get((new Foo) ! foo(true)))
    println("{}", get((new Foo) ! foo(false)))
  end
end
```

supercooldave:
This PR relaxes the type checking in the compiler dealing with unknown information (`BottomType`), which arises for instance in expressions like `Nothing` where the parameter to `Maybe[_]` is not provided nor can be inferred from the context. This PR permits such code to compile because the presence of `BottomType` does not affect the compilation.

Some previously failing tests now pass.

The key test is in `regressions/836.enc` because this PR builds directly upon the PR that fixed that issue (#841).
 
Also, fixes #792


EliasC (commented):
I see the point of this and the changes are straightforward, but I'm wondering if we're optimising for toy programs rather than "real" code? For example, the following expression now compiles (which makes sense)

```
match Nothing with
  Nothing => 42
end
```

but at the same time, the following function (still) does not compile and arguably gives a worse error message than before:

```
fun lookup[k : Eq[k], v](key : k, arr : [(k, v)]) : Maybe[v]
  var elem = Nothing
  for e <- arr do
    if e.0.eq(key) then
      elem = Just(e.1)
    end
  end
  return elem
end
```
> Not enough information to infer the type.
Try adding more type information.
In expression: 
  `Just(e.1)`
In expression: 
  `elem = Just(e.1)`

The function is fixed by adding type annotations, not where the typechecker is pointing, but where `elem` is declared. Before this PR, the typechecker would indeed fail at the declaration site.

Does it make sense to require type annotations for, say, variables and functions whose values (bodies) contain a bottom typed value?

supercooldave:
@EliasC The situation you refer to involves multiple different statements, as opposed to a single expression. Checking of bottomness doesn't span different statements.
Of the top of my head, I'm not sure how best to handle this.
Perhaps the error message could include some better hints, such as where good places to add the type annotations are.

src/types/Typechecker/Util.hs
```diff
@@ -54,7 +54,6 @@ import AST.AST as AST
 import Data.List
 import Data.Maybe
 import Text.Printf (printf)
-import Debug.Trace
```
EliasC, line 57: We will most likely need this again :)

EliasC:
What I'm proposing is to disallow `val x = Nothing` and give the error about type annotations there, but still allow e.g. `match Nothing with ...`. If you assign `Nothing` to a variable, chances are very high that you are planning to re-assign it, which will lead to an error. We should therefore catch this error as early as possible.

src/types/Typechecker/Util.hs
supercooldave: I thought we were finished with tracing. My bad.

supercooldave:
@EliasC Sounds complicated to check and awfully specific. Would `[Nothing]` be allowed? Or is the assignment of a bottom-typed value disallowed?

EliasC:
I'll be more clear:

An assignment where the right-hand side has a type that contains the bottom type should not be allowed without type annotations. Similarly, I think that closure whose body has a type that contains the bottom type should not be allowed without type annotations. All other usages of the bottom type can be allowed.

PR #847, opened 2017-07-04
Fix bug when overriding bounded polymorphic methods

EliasC:
See the added test cases for what used to fail before (and what
still correctly fails). In short, bounds for type parameters would
not get translated correctly when doing name mangling for trait
methods.


EliasC:
Assigning @supercooldave as this is related to #838 which he recently reviewed.

supercooldave:
Simple enough.

PR #849, opened 2017-07-07
Utility Programs

supercooldave:
Here are two utility programs that have been used in the course LACCP and by several project students. I translated them to Encore -- perhaps they might be useful for someone

The programs are:
* GraphGen -- generates random graphs, using for instance, preferential attachment
* FileSplitter -- splits large files -- useful in conjunction with GraphGen for generating and splitting large graphs in preparation for MapReduce

Perhaps you don't think the programs belong in this repo, or that they belong in Stress tests. Just say so, if that's your view.


supercooldave:
Oops. Somehow the libs are in there too. That's not what I want.

PR #850, opened 2017-07-07
Fixed bug 840

### LINKED ISSUE #840: Compiling Forward to Method with Arguments Generates Broken C
The following simple example
```
active class Main
  def main() : unit
    (new Worker()) ! seq(10)
  end
end

active class Worker
  def seq(s : int) : unit
    if s > 0 then
      forward (this!seq(s - 1))  -- FAILS TO COMPILE
      -- this.seq(s - 1)         -- COMPILES
    end
  end
end
```
results in broken C
```
BrokenForward_src/Worker.encore.c:185:73: error: use of undeclared identifier
      '_binop_7'; did you mean '_binop_4'?
  ..._this, NULL, _binop_7);
                  ^~~~~~~~
                  _binop_4
BrokenForward_src/Worker.encore.c:177:13: note: '_binop_4' declared here
    int64_t _binop_4 = (({ _enc__arg_s;}) - ({int64_t _literal_3 = 1; _l...
            ^
1 error generated.
```

Looking at the test directory, it seems that `forward` is not tested with methods that take arguments.

supercooldave:
Fixes a bug in the implementation of `forward`.
The variables for different branches of the body of an if statement were placed outside of the if statement resulting in errors in the generated C. These variables were moved to the right place. 

One regression test added.

Fixes #840.


kikofernandez (approved):
Looks good. Merging!

PR #851, opened 2017-07-10
JSON Parser and Decoder

supercooldave:
An implementation of a JSON parser, encoder, and decoder in Encore. The JSON parser takes a string and converts it into a data structure capturing the JSON structure, but not providing it with any interpretation. A Decoder takes that data structure and produces some Encore object. JSON decoders are constructed with an extensive combinator library, which allows one to extract fields, convert primitives, etc etc, and convert to whatever representation is suitable for your program (thereby hiding the JSON). An Encoder is used for converted into JSON.
The decoder follows the design of Elm's JSON decoder.

An extensive bunch of tests have been provided.

(PR also includes some fixes to std libs, in particular, changes to iterators so that the `next()` method does not return a `Maybe` type.)

### DIFF
--- a/modules/standard/Collections/Mutable/HashMap.enc
+++ b/modules/standard/Collections/Mutable/HashMap.enc
@@ -78,7 +78,7 @@ local class HashMapIterator[k : Hashable + Eq[k], v]
     Returns Nothing if the end of the HashMap
     has been reached.
   -}
-  def next() : Maybe[Entry[k,v]]
+  def next() : Entry[k,v]
     val current = this.current_node
 
     if this.has_next() then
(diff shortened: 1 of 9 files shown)


modules/standard/Collections/Mutable/HashMap.enc
```diff
@@ -78,16 +78,16 @@ local class HashMapIterator[k : Hashable + Eq[k], v]
...
 
     if this.has_next() then
       this.current_node = this.get_next_node()
     end
 
     match current with
-      case Just(n) => Just(n.value)
-      case Nothing => Nothing
+      case Just(n) => n.value
+      case Nothing => abort("No next element")
```
kikofernandez, line 90: Does this mean that a user cannot do `iterator.next().next()` she knows that exact number of items? If the iterator doesn't have any elements, doing `iterator.next()` crashes, doesn't it?

supercooldave: You could never do `iterator.next().next()`. The types don't work. A precondition to `next()` is that `has_next()` is true, so it shouldn't be called in a crashing situation.

The previous version of `Iterator.next()` wrapped the result in a `Maybe` type, which made it a pain to program with.

PR #856, opened 2017-07-29
Basic exception handling

sashther:
This PR adds basic exception handling to Encore, using futures to propagate exceptions during inter-actor communication.

Summary of changes/additions:
- A set of language constructs for exception handling
- A class for exception objects
- A modified version of Exceptions4C 3.0.5, a C library for exception handling by Guillermo Calvo, (http://guillermocalvo.github.io/exceptions4c)
- Changes to futures and the `get` mechanism.
- An exception context for each behaviour/processed message
- Tests for exceptions (src/tests/encore/exceptions)

## New language constructs

Parsing, IR, and codegen have been extended to provide the following language constructs:
```
  exceptionDef  ::= 'exception' exceptionType '(' exceptionType ')'
  Expr          ::= 'throw' exceptionType ['(' stringLiteral ')']
  Expr          ::= 'try' block (catchClauses ['finally' block] | 'finally' block)
  catchClauses  ::= catchClause {catchClause} | {catchClause} 'catch' block
  catchClause   ::= 'catch' [name ':'] exceptionType block
  exceptionType ::= identifierWithInitialUpperCase
```

  Note:
  - `'catch' block` catches any exception type.
  - The second `exceptionType` in `exceptionDef` is the supertype.
  - Exception types exist in the global scope. `exceptionType` is simply an identifier with an initial uppercase letter (for now).

## Exception class

- Exception.enc contains an exception class, which represents a thrown exception and allows the user to retrieve exception data.
- The stored data is message, exception type, C line and C file.
- A rethrow method is provided.
- Exceptional.enc contains an Either type for exceptions as well as a "get" function that interacts with exceptionals.

## Exceptions4C

A modified version of Exceptions4C 3.0.5 by Guillermo Calvo (http://guillermocalvo.github.io/exceptions4c) is used as a base for exception handling. Specifically the e4c_lite version is used, with the following modifications for thread safe and optimized operation (in some cases):

- Interface functions pass around an exception context.
- Frames are stored in a linked list instead of array to avoid overhead when exceptions are not used. A trade-off, since every try-expression stores a new frame to the stack.
- Changes to avoid compiler warnings:
  * added braces around try-loop and a `TRY_END` macro
  * removed some const qualifiers.
- `throw` no longer jumps/propagates when no exception frames are left. Instead, a flag "unpropagated" is set to true, and the function returns.
  * Before fulfilling, this info can be used to check whether an exception is unpropagated and if so, store it to the future.
  * Similarly, dispatch of Main can check for an unpropagated exceptions and exit the program if found.
- An exception object can be stored, alongside the e4c exception, and can be re-used by when fulfilling futures, rethrow, within catch-clauses, etc.
- Workaround for GCC bug 53119. Not platform independent!
- Removed default message from exception types.

### Futures
- The future struct has been extended with an 'exceptional' flag.
- `get` interacts with exceptions in futures, found exceptions are rethrown.
- `future_fulfil` fulfills with exception objects if unpropagated exceptions exist.

### Exception contexts
- The pony_context/encore_context now holds a pointer to an exception context.
- Calls to exception functions pass exception context as arg.
- Exception context is created and stored in the `handle_message` stack frame.
- A ptr to the exception context is stored in `pony_context`.
- Exception context ptr is stored and restored before and after context swapping (`actor_save_context`).

### Other
- Exception objects are re-used during rethrows, in catch-clauses and in future fulfills.

# Limitations/TODOs

## Exception handling with streaming methods and future chaining

This is not supported by the exception handler.

## Handling zero division and similar through OS signals

This is not supported by the exception handler. See `e4c_signal_mapping` and more in `e4c/src/e4c.c` for an example of how to do this.

## Adding exception types to the program table

Exception types are not known to Encore, but only exist in the underlying C code. This means that incorrect programs will pass typechecking, which leads to undefined behaviour. Example:
```
try
  ...
catch NonExistentExceptionType
  -- compiles, but should not! Leads to undefined behaviour (likely crash).
end
```

If the typechecker knew which exception types that existed by checking the program table, similar to how it checks classes through the program table, then this could be prevented. It would require that `exception` statements (which are top level) are somehow processed before class definitions. Each use of `exception` then adds the new exception type to the program table.

**If this is implemented:**

- Exception objects, `Exception`, currently hold references to a `struct e4c_exception_type`. This reference can be replaced with a reference to the exception type in the modifed program table.

- The compiler should also be modified to not permit identifiers with the same name as existing exception types.

- Catch clauses can be changed to also allow the syntax `'catch' identifier` to store caught exceptions. Currently `'catch' identifier ':' 'identifierWithInitialUpperCase'` is used for this purpose. If an initial upper case letter is used with `'catch' identifier`, it impossible to know whether the identifier is for the exception type or caught exception.

## `rethrow` does not work properly outside of exception frames (try-catch clauses).

The test "src/encore/exceptions/async_FAILS.enc" reproduces this bug.

`rethrow` uses the `E4C_THROW`, which will not lead to a `longjmp` in cases where there is no exception frame left. Instead the method is supposed to return, and the flag "unpropagated" of the exception context is set to true. This flag is then used in dispatch, to check whether an exception has been thrown, and to then fulfill the future with this exception. Since `rethrow` is a method, the "return" will apply to the `rethrow` method, and not the method where rethrow is invoked.

To fix, it would be best to make `throw` can be extended to operate on exception objects on top of functioning as it currently does.

## Unknown bug?

The test "encore/exceptions/async_complex_FAILS.enc" crashes.

## Passing string objects to `throw` and `exception`

The `throw` statement can only be passed string literals `throw MyException("Goodbye world!")` and not string objects or expressions `throw MyException(msgString)`. The same problem exists with the exception statement, `exception MyException("Default world!",RuntimeException)` vs. `MyException(msgString,...)`. Fixing this requires modifying typechecking so that either string objects or string literals can be passed, and in case of string objects "extract" the string literal.

## Hard coded name of Encore functions in compiler.

The name of a function "_enc__global_fun_Exceptione4c_to_exception_object"` (which allocates an exception object) is hard coded (see `e4cToExceptionObject` in `CCodeNames.hs` and `ClassDecl.hs`). Creating exception object this way is probably not very robust, and should be done in a better way.

## Making supertype optional in exception type definitions

Exception type definitions can be more expressive by making the supertype default to `RuntimeException` or similar.

`exception MyExc1` would become desugared to `exception MyExc1(RuntimeException)`

## Limited message length in thrown exceptions

Thrown exceptions have a fixed max length in C world: `struct e4c_exception{ char message[E4C_MESSAGE_SIZE]; ...};`

## Overhead during context swapping (`encore.c:actore_save_context`)

This adds a very slight overhead to running time, really only noticeable in worst-case programs, and I'm unsure of how or whether this can be optimized away. The exception data is behaviour/message local and must be correctly swapped as Encore's context swapping occurs. This is currently done in `encore.c:actor_save_context` where a pointer to the exception, residing on the stack of the current behaviour, is saved and restored before and after `assert_swap`.


kikofernandez:
Closing PR. There have been conflicts for 2 years and the merge would leave inconsistencies in the language. It was a great effort and has shown us many lessons learnt.
Thanks!

PR #860, opened 2017-10-09
Improving the ParT abstraction

kikofernandez:
This commit fixes:

* tracing futures in the ParT
* improves the ParT creating a balanced tree for the combinator `join` when it receives a `Par[Array[Par[t]]]` (I have written the `Array` as if it was a class so that it's easier to read).
* remove recursion from the `>>` combinator, which can consume the whole stack (I have seen this happening in some benchmarks that are not yet ready to be in the PR).


kikofernandez:
added test that crashes the current ParT implementation and works with this implementation

EliasC:
This code looks reasonable, although I can't comment on its correctness with respect to ParT. I'm happy to merge this later today unless there are objections.

PR #862, opened 2018-04-04
Algebraic Data Types

MicaelLoberg:
This PR extends the Encore language with ADTs, which are cool and totally sweet.
There are some examples how to use them in the tests I've added.

### DIFF
--- a/src/ir/AST/AST.hs
+++ b/src/ir/AST/AST.hs
@@ -242,6 +242,9 @@ data AdtConstructor = ADTcons {
   acfields    :: [ParamDecl]
 } deriving (Show)
 
+fieldsFromClass :: ClassDecl -> [FieldDecl]
+fieldsFromClass c@Class{cfields} = cfields
+
 instance Eq ClassDecl where
   a == b = getId (cname a) == getId (cname b)
 
--- a/src/ir/AST/AST.hs
+++ b/src/ir/AST/AST.hs
@@ -30,7 +30,8 @@ data Program = Program {
   typedefs :: [Typedef],
   functions :: [Function],
   traits :: [TraitDecl],
-  classes :: [ClassDecl]
+  classes :: [ClassDecl],
+  adts :: [AdtDecl]
 } deriving (Show)
 
 setProgramSource source p = p{source}
@@ -226,6 +227,18 @@ data ClassDecl = Class {
   ccomposition :: Maybe TraitComposition,
   cfields     :: [FieldDecl],
   cmethods    :: [MethodDecl]
+} deriving (Show)
+
+data AdtDecl = ADT {
+  ameta       :: Meta AdtDecl,
+  aname       :: Type,
+  aconstructor :: [AdtConstructor]
+} deriving (Show)
+
+data AdtConstructor = ADTcons {
+  acmeta      :: Meta AdtConstructor,
+  acname      :: Type,
+  acfields    :: [ParamDecl]
 } deriving (Show)
 
 instance Eq ClassDecl where
--- a/src/ir/AST/AST.hs
+++ b/src/ir/AST/AST.hs
@@ -232,7 +232,8 @@ data ClassDecl = Class {
 data AdtDecl = ADT {
   ameta       :: Meta AdtDecl,
   aname       :: Type,
-  aconstructor :: [AdtConstructor]
+  aconstructor :: [AdtConstructor],
+  amethods     :: [MethodDecl]
 } deriving (Show)
 
 data AdtConstructor = ADTcons {
--- a/src/ir/AST/AST.hs
+++ b/src/ir/AST/AST.hs
@@ -611,7 +611,9 @@ data Expr = Skip {emeta :: Meta Expr}
           | AdtExtractorPattern {emeta :: Meta Expr,
                                  ty :: Type,
                                  name :: Name,
-                                 arg :: Expr}
+                                 arg :: Expr,
+                                 fieldNames :: [String],
+                                 adtClassDecl :: ClassDecl}
           | ExtractorPattern {emeta :: Meta Expr,
                               ty :: Type,
                               name :: Name,
@@ -682,7 +684,8 @@ data Expr = Skip {emeta :: Meta Expr}
                  body   :: Expr}
           | Match {emeta :: Meta Expr,
                    arg :: Expr,
-                   clauses :: [MatchClause]}
+                   clauses :: [MatchClause],
+                   adtMatch::Bool}
           | Borrow {emeta  :: Meta Expr,
                     target :: Expr,
                     name   :: Name,
--- a/src/ir/AST/AST.hs
+++ b/src/ir/AST/AST.hs
@@ -31,7 +31,8 @@ data Program = Program {
   functions :: [Function],
   traits :: [TraitDecl],
   classes :: [ClassDecl],
-  adts :: [AdtDecl]
+  adts :: [AdtDecl],
+  adtCons :: [AdtConstructor]
 } deriving (Show)
 
 setProgramSource source p = p{source}
@@ -230,16 +231,20 @@ data ClassDecl = Class {
 } deriving (Show)
 
 data AdtDecl = ADT {
-  ameta       :: Meta AdtDecl,
-  aname       :: Type,
+  ameta        :: Meta AdtDecl,
+  aname        :: Type,
   aconstructor :: [AdtConstructor],
-  amethods     :: [MethodDecl]
+  amethods     :: [MethodDecl],
+  identity     :: String
 } deriving (Show)
 
 data AdtConstructor = ADTcons {
-  acmeta      :: Meta AdtConstructor,
-  acname      :: Type,
-  acfields    :: [ParamDecl]
+  acmeta         :: Meta AdtConstructor,
+  acname         :: Type,
+  acfields       :: [ParamDecl],
+  acomposition   :: TraitComposition,
+  acmethods      :: [MethodDecl],
+  parentIdentity :: String
 } deriving (Show)
 
 fieldsFromClass :: ClassDecl -> [FieldDecl]
--- a/src/ir/AST/AST.hs
+++ b/src/ir/AST/AST.hs
@@ -684,8 +684,7 @@ data Expr = Skip {emeta :: Meta Expr}
                  body   :: Expr}
           | Match {emeta :: Meta Expr,
                    arg :: Expr,
-                   clauses :: [MatchClause],
-                   adtMatch::Bool}
+                   clauses :: [MatchClause]}
           | Borrow {emeta  :: Meta Expr,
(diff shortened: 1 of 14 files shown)


supercooldave:
Could be more tests, but given that Micael has moved on, I think this is worth keeping and fixing in the future.

EliasC:
I have a review of this PR that I did while offline yesterday. Will post it here for posterity tomorrow.

EliasC:
I wrote this review a day before it was merged. @MicaelLoberg is not expected to address these things (although if he wants to, please tell us below!). This is just the high-level comments. I will provide the code review later (for whomever is going to fix it).

- `data` should be a reserved keyword

- It's a bit annoying that a data constructor without arguments
  still needs parentheses:

  ```
  data MyBool
  case True() : MyBool
  case False : MyBool -- Parsing error: needs parentheses

  active class Main
    def main() : unit
      val t : MyBool = True -- fails: True is a function of type () -> MyBool
    end
  end
  ```

  This comes down to design decisions, and could probably be left
  as is for now.

- Providing a non existing data type for a case crashes the compiler:

  ```
  data MyBool
  case True() : MyBol
  ```
  > encorec: "adt.enc" (line 2, column 1): ADT "MyBol" not defined

  This looks like an attempt to do "type checking" in the
  desugarer, which is too early because we have no type
  information yet. The Haskell function `error` should _only_ be
  used for compiler bugs, i.e. things that should never happen.
  Blindly passing the type and let the type checker sort it out
  should be enough.

- Related to the above, using a trait or class type as the type of
  a case crashes the compiler:

  ```
  trait MyBool
  case True() : MyBool
  ```
  > encorec: "adt.enc" (line 10, column 1): ADT "Expr" not defined

- The parser allows trait compositions as the type of data cases,
  but this crashes the compiler:

  ```
  data MyBool
  data Foo
  case True() : MyBool + Foo -- Causes a crash!
  ```
  > encorec: src/ir/AST/Desugarer.hs:274:39-75: Non-exhaustive patterns in record update

  The simplest solution is of course to only allow a single
  ref-type (although it is exciting to think about what it would
  mean to have a case that belongs to two different datatypes!).

- Type inference for polymorphic data should be improved:

  ```
  data List[t]
  case Nil[t]() : List[t]
  case Cons[t](hd : t, tl : List[t]) : List[t]

  active class Main
    def main() : unit
      var l = Cons(1, Cons(2, Cons(3, Nil[int]())))
    end
  end
  ```

  Ideally, we wouldn't have to provide the type parameter to `Nil`
  I think you could leave this for future work.

- You can currently define methods in data declarations. It would
  make sense to also be able to `require` methods (which must then
  be provided by the cases):

  ```
  data Expr
    require eval() : int
  end

  case Add(l : Expr, r : Expr) : Expr
    def eval() : int
      this.l.eval() + this.r.eval()
    end
  end
  ```

  Since the data types are implemented as traits, this should more
  or less be a matter of telling the parser to allow (method)
  requirements as well.

- The parser is a bit eager to abort parsing a block data
  definition:

  ```
  data Foo
    def foo() : unit
      1 +
    end
  end
  ```
  > "adt.enc" (line 2, column 3):
  > unexpected 'd'

  The error should of course be on the next line, where is a
  missing expression.

- Related to the above, it isn't possible to override methods in
  cases:

  ```
  data Expr
    def eval() : int
      return -1
    end
  end

  case Add(l : Expr, r : Expr) : Expr
    def eval() : int
      this.l.eval() + this.r.eval()
    end
  end
  ```

  > "adt.enc" (line 1, column 1)
  > Duplicate declaration of method 'eval'
  > In class 'Add'

  (note also that the position in the error message is off)

  Manually desugaring into the corresponding trait/class pair does
  work, which suggests that the desugaring process for ADTs could
  be simplified:

  ```
  read trait Expr
    def eval() : int
      return -1
    end
  end

  class Add : Expr(l, r)
    val l : Expr
    val r : Expr
    def eval() : int
      this.l.eval() + this.r.eval()
    end
  end
  ```

src/ir/AST/AST.hs
```diff
@@ -228,6 +230,26 @@ data ClassDecl = Class {
...
 
+data AdtDecl = ADT {
+  ameta        :: Meta AdtDecl,
+  aname        :: Type,
+  aconstructor :: [AdtConstructor],
+  amethods     :: [MethodDecl],
+  identity     :: String
+} deriving (Show)
+
+data AdtConstructor = ADTcons {
```
EliasC, line 241: I would call these `AdtCase` and `ADTCase` respectively.

src/ir/AST/AST.hs
```diff
@@ -228,6 +230,26 @@ data ClassDecl = Class {
...
+  acmeta         :: Meta AdtConstructor,
+  acname         :: Type,
+  acfields       :: [ParamDecl],
+  acomposition   :: TraitComposition,
```
EliasC, line 245: Unless we want the ability to write compositions here, this should most likely be a `Type`.

src/ir/AST/AST.hs
```diff
@@ -228,6 +230,26 @@ data ClassDecl = Class {
...
+  acmethods      :: [MethodDecl],
+  parentIdentity :: String
```
EliasC, line 247: Is this the name of the `data` declaration? If so, it should be a `Type`.

src/ir/AST/AST.hs
```diff
@@ -255,6 +277,20 @@ instance HasMeta ClassDecl where
...
+    setType _ _ =
+        error "AST.hs: Cannot set the type of an ADT"
+    showWithKind ADT{aname} = "data '" ++ getId aname ++ "'"
+
+instance HasMeta AdtConstructor where
+    getMeta = acmeta
+    setMeta ac m = ac{acmeta = m}
+    setType _ _ =
+        error "AST.hs: Cannot set the type of an ADT"
+    showWithKind ADTcons{acname} = "data '" ++ getId acname ++ "'" --TODO: ?
```
EliasC, line 292: It doesn't make sense to print both `data` and `case` declarations as "data". In other words, the TODO is well placed =)

src/types/Types.hs
```diff
@@ -809,6 +883,19 @@ isRefType ty
 isUnresolved Type{inner = Unresolved{}} = True
 isUnresolved _ = False
 
+getAdtTag Type{inner = ClassType{adtTag}} = adtTag
+getAdtTag _ = 0
```
EliasC, line 887: If this function is only supposed to be called on class types, it's better to throw an error here!

src/types/Types.hs
```diff
@@ -794,6 +866,8 @@ isRefAtomType Type{inner = Unresolved {}} = True
 isRefAtomType Type{inner = TraitType {}} = True
 isRefAtomType Type{inner = AbstractTraitType {}} = True
 isRefAtomType Type{inner = ClassType {}} = True
+isRefAtomType Type{inner = AdtType {}} = True -- TODO: Ask someone about this
+isRefAtomType Type{inner = AdtConsType {}} = True -- TODO: Ask someone about this
```
EliasC, line 870: Did you ask someone? =)

src/types/Types.hs
```diff
@@ -502,6 +517,8 @@ showWithKind ty = kind (inner ty) ++ " " ++ show ty
                                          else if isSharedSingleType (typ ty)
                                          then "shared class type"
                                          else "class type"
+    kind AdtType{}                     = "ADT type"
```
EliasC, line 520: Shouldn't this be "abstract data type"?

src/types/Typechecker/Typechecker.hs
```diff
@@ -1100,6 +1115,75 @@ instance Checkable Expr where
...
+        doGetAdtPatternVars pt fcall@(FunctionCall {qname, args = []}) = do
+          {-header <- findMethod pt (qnlocal qname)-}
+          {-let hType = htype header-}
+              {-extractedType = getResultType hType-}
+          {-unless (isUnitType extractedType) $ do-}
+                 {-let expectedLength = if isTupleType extractedType-}
+                                      {-then length (getArgTypes extractedType)-}
+                                      {-else 1-}
+                 {-tcError $ PatternArityMismatchError (qnlocal qname)-}
+                           {-expectedLength 0-}
```
EliasC, line 1143: Dead code

src/types/Typechecker/Typechecker.hs
```diff
@@ -1100,6 +1115,75 @@ instance Checkable Expr where
...
+        doGetAdtPatternVars pt va@(VarAccess {qname}) = do
+          when (isThisAccess va) $
+            tcError ThisReassignmentError
+          return [(qnlocal qname, pt)]
+
+        {-doGetAdtPatternVars pt mcp@(MaybeValue{mdt = JustData {e}})-}
+            {-| isMaybeType pt =-}
+                {-let innerType = getResultType pt-}
+                {-in getAdtPatternVars innerType e-}
+            {-| otherwise = tcError $ PatternTypeMismatchError mcp pt-}
```
EliasC, line 1132: Dead code

src/parser/Parser/Parser.hs
```diff
@@ -750,6 +758,131 @@ partitionClassAttributes = partitionClassAttributes' [] []
...
+--
+--    lineDecl ameta = do
+--      name <- lookAhead upperChar >> identifier
+--      params <- optionalTypeParameters
+--      return ADT{ameta
+--                ,aname = setRefNamespace emptyNamespace $ adtType name params
+--                ,aconstructor = []
+--                ,amethods = []
+--                ,identity = name
+--                }
```
EliasC, line 831: Lots of dead code!

src/ir/AST/AST.hs
```diff
@@ -586,6 +622,12 @@ data Expr = Skip {emeta :: Meta Expr}
                          args :: Arguments}
           | Optional {emeta :: Meta Expr,
                       optTag :: OptionalPathComponent}
+          | AdtExtractorPattern {emeta :: Meta Expr,
+                                 ty :: Type,
+                                 name :: Name,
+                                 arg :: Expr,
+                                 fieldNames :: [String],
+                                 adtClassDecl :: ClassDecl}
```
EliasC, line 630: Does this have to be a `ClassDecl`, or could it be a type that is looked up when needed?

src/ir/AST/AST.hs
```diff
@@ -228,6 +230,26 @@ data ClassDecl = Class {
   cmethods    :: [MethodDecl]
 } deriving (Show)
 
+data AdtDecl = ADT {
+  ameta        :: Meta AdtDecl,
+  aname        :: Type,
+  aconstructor :: [AdtConstructor],
+  amethods     :: [MethodDecl],
+  identity     :: String
```
EliasC, line 238: Isn't this information already in `aname`?

EliasC (commented):
A few comments on the code.

PR #867, opened 2018-06-29
Upgrade GHC from 7.10.3 to 8.0.2 with stackage LTS

albertnetymk:
This commit upgrades `pretty` so that the upcoming error message feature
could be merged. The only "manual" change is in `stack.yaml`, while the
rest is instructed by the compiler, as trying to build the project.

### DIFF
--- a/encore.cabal
+++ b/encore.cabal
@@ -30,16 +30,16 @@ executable encorec
                   , CPP
                   , ScopedTypeVariables
   build-depends: MissingH
-               , base >=4.7 && <4.9
+               , base
                , containers >= 0.5.5.1
-               , directory >=1.2 && <1.3
+               , directory
                , hashable
                , hpc
                , mtl ==2.2.*
-               , megaparsec >= 5.1.2
+               , megaparsec >= 5.1.2 && < 5.2
                , semigroups
-               , pretty >=1.1 && <1.2
-               , process >=1.2 && <1.3
+               , pretty
+               , process
                , template-haskell
                , text >=1.1
                , ghc >= 7.10
@@ -48,5 +48,5 @@ executable encorec
                , boxes
                , filepath
   hs-source-dirs:      src/back src/front src/ir src/opt src/parser src/types
-  ghc-options: -Werror
+  ghc-options: -Werror -fmax-pmcheck-iterations=10000000
   default-language:    Haskell2010
--- a/stack.yaml
+++ b/stack.yaml
@@ -4,4 +4,4 @@ packages:
 - '.'
 extra-deps:
 - megaparsec-5.1.2
-resolver: lts-6.0
+resolver: lts-9.17
--- a/src/types/Typechecker/Capturechecker.hs
+++ b/src/types/Typechecker/Capturechecker.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE ConstrainedClassMethods #-}
 {-|
 
 Capturechecks an "AST.AST" and produces the same tree, extended


encore.cabal
```diff
@@ -48,5 +48,5 @@ executable encorec
                , boxes
                , filepath
   hs-source-dirs:      src/back src/front src/ir src/opt src/parser src/types
-  ghc-options: -Werror
+  ghc-options: -Werror -fmax-pmcheck-iterations=10000000
```
EliasC, line 51: Can I ask what this is for? Not against it, just curious!

EliasC:
This looks very reasonable. Merging in a day or two unless anyone disagrees.

encore.cabal
albertnetymk: Encountered some warning without it; more at https://ghc.haskell.org/trac/ghc/ticket/11822

Thizizmyname: Between GHC 7.10 and 8.0.2 they lowered the limit, we raises the limit back up to handle the warning we otherwise get from putChildren.

.../encore/src/ir/AST/Util.hs:145:1: warning:
   Pattern match checker exceeded (2000000) iterations in
   an equation for ‘putChildren’. (Use -fmax-pmcheck-iterations=n
   to set the maximun number of iterations to n)

EliasC:
Since we're already changing `encore.cabal`, does it make sense to do something about this warning?

```
Warning: The following modules should be added to exposed-modules or other-modules in /Users/Elias/Dropbox/PhD/encore/encore.cabal:
             - In encorec component:
                 AST.AST
                 AST.Desugarer
                 AST.Meta
                 AST.PrettyPrinter
                 AST.Util
                 CCode.Main
                 CCode.PrettyCCode
                 CodeGen.CCodeNames
                 CodeGen.ClassDecl
                 CodeGen.ClassTable
                 CodeGen.Closure
                 CodeGen.Context
                 CodeGen.DTrace
                 CodeGen.Expr
                 CodeGen.Function
                 CodeGen.GC
                 CodeGen.Header
                 CodeGen.Main
                 CodeGen.MethodDecl
                 CodeGen.Preprocessor
                 CodeGen.Program
                 CodeGen.Shared
                 CodeGen.Trace
                 CodeGen.Type
                 CodeGen.Typeclasses
                 Identifiers
                 Literate
                 Makefile
                 ModuleExpander
                 Optimizer.Optimizer
                 Parser.Parser
                 SystemUtils
                 Typechecker.Capturechecker
                 Typechecker.Environment
                 Typechecker.Prechecker
                 Typechecker.TypeError
                 Typechecker.Typechecker
                 Typechecker.Util
                 Types
                 Utils
         
         Missing modules in the cabal file are likely to cause undefined reference errors from the linker, along with other problems.
```

encore.cabal
```diff
@@ -30,16 +30,16 @@ executable encorec
...
   build-depends: MissingH
-               , base >=4.7 && <4.9
+               , base
                , containers >= 0.5.5.1
-               , directory >=1.2 && <1.3
+               , directory
                , hashable
                , hpc
                , mtl ==2.2.*
                , megaparsec >= 5.1.2
```
Thizizmyname: I would suggest setting this dependency to megaparsec < 5.2
Since 5.3.1 is included as default by the lts, and 5.2 and above breaks compatibility.

That will also make stack solver not recommend deleting the extra-dep in stack.yaml
(Note, if using an unsupported cabal-install *stack solver* will error out due stack can't parse the output but it will compile fine. running with supported cabal-install-1.24 which is in lts-9.17 will successfully determine a build plan)

Thizizmyname: I would suggest setting megaparsecs' dependency to megaparsec >= 5.1.2 && < 5.2
Since 5.3.1 is included as default by the lts, and 5.2 and above breaks compatibility.

That will also make stack solver not recommend deleting the extra-dep in stack.yaml
(Note, if using an unsupported cabal-install stack solver will error out due stack can't parse the output but it will compile fine. running with supported cabal-install-1.24 which is in lts-9.17 will successfully determine a build plan)

stack.yaml
```diff
@@ -4,4 +4,4 @@ packages:
 - '.'
 extra-deps:
 - megaparsec-5.1.2
-resolver: lts-6.0
+resolver: lts-9.17
```
Thizizmyname, line 7: Latest lts for GHC 8.0.2 is 9.21.
Any reason why that is not chosen?

albertnetymk:
> does it make sense to do something about this warning

Even though we are touching that file, the solution would be orthogonal to the upgrading this commit addresses, so I suggest to put that to its own commit/PR.

> Any reason why that is not chosen?

No particular reason; any LTS along with GHC 8.0.2 suffices. We are not trying to chase after the latest LTS, which evolves too frequently.

PR #868, opened 2018-07-16
Extended `encore.cabal` with `other-modules`

EliasC:
When compiling, stack would warn about used modules not being
listed in `encore.cabal`. This commit adds them as suggested by
stack.


albertnetymk:
Will merge tomorrow if no objections.

PR #879, opened 2019-05-08
For-Comprehension

ElieOaks:
**For - Comprehension**
A new kind of for-comprehension is being implemented into Encore that replaces the old for-loop.
For-comprehension can:
-- act as a function call :
```
var x = for y <- someList do
              y
            end
```

-- iterate over itterable objects, not just Arrays and Ranges:
```
for x <- linkedList do
        --body
end

```
-- iterate over multiple data structures in one call:
```
for x <- [1, 2, 3], y <- ["a", "b", "c"] do
        -- body
end
```
-- be passed as an argument
```
foo( for y <- someList do
            y
        end)
```
The new for-comprehension does not allow:
-- Different kinds of data structures in header
```
for x <- linkedList, y <- ["a", "b", "c"] do
        -- body
end
```
-- Range to return a collection. It will compile, but x will be of type unit. Therefor ranges only desugar into `foreach`.
```
var x = for y <- [0 .. 10] do
              y
            end
```
-- A return in the body of the loop to return something outside of it. The bellow example will not return 1.
```
fun foo() : int
  for x <- [1, 2, 3]
    return x
  end
end
```
Rather it is equivalent to: 
```
fun foo() : int
  var returnCollection  for x <- [1, 2, 3]
                                        return x
                                   end
 return returnCollection
end
```
Which of course would break the typing rules, as it returns an array, and not an int.
**Under the hood:**
The for-comprehension is pure syntactic sugar, and is desugared into combination of calls to the methods ` map`, `flatMap`, `foreach`, and if the loop contains `break`, into `maybeForeach`. These methods can be found in some of the collections in the standard libraries, such as Array and LinkedList.
the methods map and flatMap are used in combination, when a return collection is expected. Otherwise the method foreach is used, as it's loop has only side effects. 
```
var x = for y <- someList, z <- someListTwo do
              y + z
            end
```
is desugaraed into: 
`var x = someListTwo.flatMap(someList.map(y))`

_______________________________________________

```
for y <- [0 .. 10] do
         print(y)
end
```
is desugaraed into: 
` [0 .. 10].foreach(print(y))`

________________________________________________

```
for y <- [0 .. 10] do
         print(y)
         if y > 9 then
            break
         end
end
```
is desugaraed into: 
```
 [0 .. 10].maybeForeach(
  print(y)
  if y > 9 then
     return Nothing
  end
  Just(())
  )

```
________________________________________________
**Added compilation steps**
The desugaring of for-comprehension depends upon typing information, and is therefore a new step of compilation called TypedDesugarer, and occurs before the optimizations. As the desusaring process breaks down the For-Expr and builds up a sub-AST of new Expr, the AST needs to be re-type checked, and re-capture checked. Therefore the compilation includes Typed Desugaring, as well as Re - Type checking, and Re - Capture checking. When modular type checking, and capure checking becomes available, these compilation steps can hopefully be removed. 

**Problem with re - type checking**
Type checking twice is not possible due to the type checking process of Match clauses. When Match is type checked, it changes the Match-clause so that it violates its own typing rules and cannot be type checked again without throwing an Error. A hack has been created so that programs with for-comprehension can compile and run, but is currently incompatible with the match - type checker. This fork can be merges only when the match type checking problem has been solved.

**Arrays**
Unlike classes such LinkedList in the standard Library, there is a list of functions, rather than methods that can be applied to arrays. Therefore for-comprehensions that are over arrays are desugared into function calls ` map`, `flatMap`, `foreach`, and if the loop contains `break`, into `maybeForeach`, rather than into method calls by the same name. 

**Other changes**
This PR has also had to incorporate the PR: Ability to compile pony runtime with gcc 7.x #865. As the latest ubuntu, which this was developed on, runs on gcc 7. 
It also desugars ranges into a class RRange, that can be found in the module-file Data. This was in order to have the ability to add methods. As range is a protected word in the compiler, the class has to be named something else, hence the double R:s. As ranges are only used together with for-loops / for-comprehensions, this change is not expected to affect anything else.


modules/standard/Collections/Mutable/Functor.enc
```diff
@@ -0,0 +1,10 @@
+-- JOY for-comprehension
```
kikofernandez: It is nice to have the name, but I would remove this line :)

src/front/TopLevel.hs
```diff
@@ -341,6 +341,14 @@ main =
        verbose options "== Optimizing =="
        let optimizedTable = fmap optimizeProgram capturecheckedTable
 
+       -- JOY for-comprehension
```
kikofernandez: be careful with these ones.

src/ir/AST/AST.hs
```diff
@@ -684,11 +691,10 @@ data Expr = Skip {emeta :: Meta Expr}
                     name :: Name,
                     times :: Expr,
                     body :: Expr}
+          -- JOY for-comprehension
```
kikofernandez: no need to annotate your stuff :)

src/ir/AST/AST.hs
```diff
@@ -600,6 +600,13 @@ data VarDecl =
              varType :: Type}
   deriving(Eq, Show)
 
+-- JOY for-comprehension
+data ForSource =
+     ForSource { forVar :: Name,
```
kikofernandez: following Encore compiler convention, this should be `name`, and the `Maybe Type` should be `ty`

src/ir/AST/Desugarer.hs
```diff
@@ -389,7 +389,8 @@ desugar FunctionCall{emeta, qname = QName{qnlocal = Name "exit"}
     Exit emeta args
 
 -- Abort
-desugar FunctionCall{emeta, qname=QName{qnlocal=Name "abort"} , args=[msg]} =
+desugar FunctionCall{emeta, qname = QName{qnlocal = Name "abort"}
```
kikofernandez, line 392: ?

src/ir/AST/PrettyPrinter.hs
```diff
@@ -369,15 +369,14 @@ ppExpr Repeat {name, times, body} =
...
     "end"
-ppExpr For {name, step = IntLiteral{intLit = 1}, src, body} =
-    "for" <+> ppName name <+> "<-" <+> ppExpr src <+> "do" $+$
-         indent (ppBody body) $+$
-    "end"
-ppExpr For {name, step, src, body} =
-    "for" <+> ppName name <+> "<-" <+> ppExpr src <+>
-    "by" <+> ppExpr step <+> "do" $+$
-         indent (ppBody body) $+$
+-- JOY for-comprehension
```
kikofernandez: remove name

src/ir/AST/Util.hs
```diff
@@ -93,7 +93,11 @@ getChildren Unless {cond, thn} = [cond, thn]
 getChildren While {cond, body} = [cond, body]
 getChildren DoWhile {cond, body} = [cond, body]
 getChildren Repeat {name, times, body} = [times, body]
-getChildren For {name, step, src, body} = [step, src, body]
+-- JOY for-comprehension
+getChildren For {sources, body} = body : concatMap getChildrenFor' sources
+ where
+    getChildrenFor' ForSource {collection} = [collection]
+-- getChildren For {name, step, src, body} = [step, src, body]
```
kikofernandez: remove line if not used

src/ir/AST/Util.hs
```diff
@@ -93,7 +93,11 @@ getChildren Unless {cond, thn} = [cond, thn]
 getChildren While {cond, body} = [cond, body]
 getChildren DoWhile {cond, body} = [cond, body]
 getChildren Repeat {name, times, body} = [times, body]
-getChildren For {name, step, src, body} = [step, src, body]
+-- JOY for-comprehension
+getChildren For {sources, body} = body : concatMap getChildrenFor' sources
```
kikofernandez, line 96: This looks fine. I do not recall if the order of the children matter. 
If you can check this and see that it does, you should first process the sources and concat the body

src/ir/AST/Util.hs
```diff
@@ -173,7 +177,10 @@ putChildren [cond, thn] e@(Unless {}) = e{cond = cond, thn = thn}
 putChildren [cond, body] e@(While {}) = e{cond = cond, body = body}
 putChildren [cond, body] e@(DoWhile {}) = e{cond = cond, body = body}
 putChildren [times, body] e@(Repeat {}) = e{times = times, body = body}
-putChildren [step, src, body] e@(For {}) = e{step = step, src = src, body = body}
+-- JOY for-comprehension
+putChildren (body:sourceList) e@(For {sources}) =
+  e{body = body, sources = zipWith (\collec s -> s{collection = collec}) sourceList sources}
+--putChildren [step, src, body] e@(For {}) = e{step = step, src = src, body = body}
```
kikofernandez: remove this line if not used

src/ir/AST/Util.hs
```diff
@@ -173,7 +177,10 @@ putChildren [cond, thn] e@(Unless {}) = e{cond = cond, thn = thn}
 putChildren [cond, body] e@(While {}) = e{cond = cond, body = body}
 putChildren [cond, body] e@(DoWhile {}) = e{cond = cond, body = body}
 putChildren [times, body] e@(Repeat {}) = e{times = times, body = body}
-putChildren [step, src, body] e@(For {}) = e{step = step, src = src, body = body}
+-- JOY for-comprehension
+putChildren (body:sourceList) e@(For {sources}) =
+  e{body = body, sources = zipWith (\collec s -> s{collection = collec}) sourceList sources}
```
kikofernandez: I am not sure this is right... Based on the other ASTs, the other AST never use the value inside the AST, e.g., look at the above `Repeat` AST and check how we never access anything on the inside, we just "update" them.
Based on your `getChildren` implementation, I believe that the `sourceList` is a list of `collection` which needs to be wrapped around a `ForSource` (I think...)

src/ir/AST/Util.hs
```diff
@@ -443,8 +450,13 @@ freeVariables bound expr = List.nub $ freeVariables' bound expr
...
             let xs = map (qLocal . varName) vars
             in (freeVariables' bound expr ++ free, xs ++ bound)
-    freeVariables' bound e@For{name, step, src, body} =
-      freeVariables' (qLocal name:bound) =<< getChildren e
+    -- JOY for-comprehension not sure what to do here
+    freeVariables' bound e@For{sources, body} =
+      freeVariables' (getForVar++bound) =<< getChildren e
+      where
+        getForVar = map (\ForSource{forVar, collection} -> qLocal forVar) sources
+    {-freeVariables' bound e@For{name, step, src, body} =
```
kikofernandez: remove this lines

src/ir/AST/Util.hs
```diff
@@ -475,8 +487,13 @@ mark asParent s@Let{body, decls} =
...
     markDecl (n, e) = (n, markAsExpr e)
 mark asParent s@While{cond, body} = asParent s{cond=markAsExpr cond, body=markAsStat body}
-mark asParent s@For{step, src, body} =
-  asParent s{step=markAsExpr step, src=markAsExpr src, body=markAsStat body}
+-- JOY for-comprehension, doesn't work, type matching is wrong.
+mark asParent s@For{sources, body} =
+  asParent s{sources = map markAsForSource sources, body=markAsStat body}
+  where
+    markAsForSource ForSource{forVar, forVarType, collection} = ForSource {forVar, forVarType, collection = markAsExpr collection}
+{-mark asParent s@For{step, src, body} =
```
kikofernandez: remove this

kikofernandez: try to stick to the 80 characters width rule, I think this is mostly used in the whole code base

src/opt/Optimizer/Optimizer.hs
```diff
@@ -49,7 +54,7 @@ optimizeProgram p@(Program{classes, traits, functions}) =
 -- | The functions in this list will be performed in order during optimization
 optimizerPasses :: [Expr -> Expr]
 optimizerPasses = [constantFolding, sugarPrintedStrings, tupleMaybeIdComparison,
-                   dropBorrowBlocks, forwardGeneral]
+                   dropBorrowBlocks, forwardGeneral] ++ [forDesugared]
```
kikofernandez:
```
[ ..., forwardGeneral, forDesugared]
```

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+-- end
+--
+-- into listA.flatMap(listB.flatMap(listC.map(fun)))
+--
+-- Credit: kaeluka for the use of foldl1 and zipWith in this manner
+forDesugared = extend forDesugared'
+  where
+    forDesugared' :: Expr -> Expr
+    forDesugared' e@For{emeta, sources, body} =
+      let
```
kikofernandez: I would try to continue writing in the same Encore codebase style, e.g., 
```
let n = ...
     x = ...
in zzz
```

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+        n = length sources
+        callNameList = if AST.AST.isCaptured e
+                      then replicate n (Name "foreach")
+                      else replicate (n-1) (Name "flatMap") ++ [Name "map"]
+        revSources = reverse sources
+        elemType = bodyType body
+        forprettyprint = nestCalls emeta callNameList sources body elemType
+      in
+        trace (show (ppExpr forprettyprint)) forprettyprint
```
kikofernandez: do not forget to remove the trace once you have something stable

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+    forDesugared' e = e
+
+nestCalls :: Meta.Meta Expr -> [Name] -> [ForSource] -> Expr -> Type -> Expr -- nested MethodCalls and FunctionCalls
```
kikofernandez: comments better above the typing

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+nestCalls meta (name:_) (fs:[]) body elemType = intoCall meta name fs body elemType
+nestCalls meta (name:restOfNames) (fs:restFS) body elemType =
+  let nestedCall = intoCall meta name fs body elemType
+  in nestCalls meta restOfNames restFS nestedCall elemType
+
+intoCall :: Meta.Meta Expr -> Name -> ForSource -> Expr -> Type -> Expr -- MethodCall or FunctionCall
+intoCall met callName ForSource{forVar, forVarType, collection} bodyOrMethodCall elemType =
+  if isRefType (getType collection)
+  then let
```
kikofernandez: `let x = ...`

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+        param = [intoParam met Val forVar forVarType]
+        arguments = [intoClosure met param Nothing bodyOrMethodCall]
+       in
```
kikofernandez: `in xxx`

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+        intoMethodCall met [elemType] collection callName arguments
+   else let
```
kikofernandez: same

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+        param = [intoParam met Val forVar forVarType]
+        arguments = [intoClosure met param Nothing bodyOrMethodCall] ++ [collection]
+       in
```
kikofernandez: same

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+        intoFunctionCall met [(fromMaybe intType forVarType), elemType] callName arguments
+
+-- helper functions
+bodyType body = getType body
```
kikofernandez: this function seems to be only used once. It should be removed if that is the case :)

src/opt/Optimizer/Optimizer.hs
```diff
@@ -175,3 +180,90 @@ forwardGeneral = extend forwardGeneral'
...
+
+-- Maybe these should be in a kind of Util file, or in AST?
+intoClosure meta parameters mty body =
```
kikofernandez: I have mixed feelings about this functions. They would probably be useful for the Typechecker and others, but we are not using them there, we write the algebraic data type as it is. To maintain uniformity in the compiler, these ones should not be present. We can discuss this offline

src/parser/Parser/Parser.hs
```diff
@@ -1443,18 +1443,20 @@ expr = notFollowedBy nl >>
         cond <- expression
         reserved "then"
         return $ \thn -> Unless{emeta, cond, thn}
-
-      for = blockedConstruct $ do
+      -- JOY for-comprehension
```
kikofernandez: remove the name

src/parser/Parser/Parser.hs
```diff
@@ -1443,18 +1443,20 @@ expr = notFollowedBy nl >>
...
-        step <- option (IntLiteral stepMeta 1)
-                       (do {reserved "by"; expression})
+        sources <- option [] $ (commaSep getForSource)
         reserved "do"
-        return $ \body -> For{emeta, name, src, step, body}
+        return $ \body -> For{emeta, sources, body}
+        where
+          getForSource = do
+             forVar <- Name <$> identifier
+             let forVarType = Nothing
```
kikofernandez: you can remove this line and add directly the `Nothing` in the `ForSource` a few lines below.

src/types/Typechecker/Typechecker.hs
```diff
@@ -1722,33 +1722,43 @@ instance Checkable Expr where
...
-           bodyTyped <- typecheckBody elementType body
-           return $ setType unitType for{step = stepTyped
-                                        ,src  = srcTyped
-                                        ,body = bodyTyped}
-        where
-          addIteratorVariable ty = extendEnvironmentImmutable [(name, ty)]
-          typecheckBody ty = local (addIteratorVariable ty) . typecheck
+
+    -- JOY for-comprehension
+    -- returnType is unittye, is changed during the second typechecking when it has been desugaraed into a method call
```
kikofernandez: - too long comment. break down into multiple lines if bigger than 80 char threshold
- remove name
- if you can, try to add the typing rules from the comment of the previous for-comprehension implementation

src/types/Typechecker/Typechecker.hs
```diff
@@ -1722,33 +1722,43 @@ instance Checkable Expr where
...
+      where
+        typeCheckSources :: [ForSource] -> TypecheckM [ForSource]
+        typeCheckSources sourceList =  do
+          typedSources <- mapM typeCheckSource sourceList
+          return typedSources
+        typeCheckSource fors@(ForSource{forVarType, collection}) = do
+            collectionTyped <- doTypecheck collection
+            let collectionType = AST.getType collectionTyped
+            let forVarType = return $ getInnerType collectionType
+            return fors{forVarType = forVarType,
```
kikofernandez: for consistency, the Haskell style does:
```
fors{... = ...
      ,... = ...}
```

src/types/Typechecker/Typechecker.hs
```diff
@@ -1722,33 +1722,43 @@ instance Checkable Expr where
...
+    doTypecheck for@(For {sources, body}) = do
+      sourcesTyped <- typeCheckSources sources
+      forVarList <- getForVarTypeList sources
+      bodyTyped <- typecheckBody forVarList body
+      return $ setType unitType for{sources = sourcesTyped,
+                                      body = bodyTyped}
+      where
+        typeCheckSources :: [ForSource] -> TypecheckM [ForSource]
+        typeCheckSources sourceList =  do
+          typedSources <- mapM typeCheckSource sourceList
```
kikofernandez: if the return type is the result of the `mapM`, you move the line `mapM typCheckSource sourceList` to the line above, removing the `do` and the `return typeSources`

kikofernandez: If you do the comment above, maybe line 1729 is really a `mapM typeCheckSource sourceList`, and you have to maintain one less function

src/types/Typechecker/Typechecker.hs
```diff
@@ -1722,33 +1722,43 @@ instance Checkable Expr where
...
+          return typedSources
+        typeCheckSource fors@(ForSource{forVarType, collection}) = do
+            collectionTyped <- doTypecheck collection
+            let collectionType = AST.getType collectionTyped
+            let forVarType = return $ getInnerType collectionType
+            return fors{forVarType = forVarType,
+                        collection = setType collectionType collectionTyped}
+
+        getForVarTypeList sourceList = mapM getForVarType sourceList
+        getForVarType ForSource{forVar, collection} = do
```
kikofernandez: I would add the typing to this function, I think it can serve as documentation giving that returning a tuple is not the most common thing for the compiler

src/types/Typechecker/Typechecker.hs
```diff
@@ -1722,33 +1722,43 @@ instance Checkable Expr where
...
+             pushError collection $ NonIterableError collectionType
+          let forVarType = getInnerType collectionType
+          return (forVar, forVarType)
+
+        getInnerType collectionType
+         | isArrayType collectionType = getResultType collectionType
+         | isRangeType collectionType = intType
+         | otherwise = head $ getTypeParameters collectionType
+
+        addIteratorVariable forVarList = extendEnvironmentImmutable forVarList
```
kikofernandez: this line says that `addIteratorVariable = extendEnvironmentImmutable`. Hence, I would remove this equivalence since it is just a renaming and use the `extendXXXX`

src/types/Typechecker/Typechecker.hs
```diff
@@ -1722,33 +1722,43 @@ instance Checkable Expr where
...
+        getForVarTypeList sourceList = mapM getForVarType sourceList
+        getForVarType ForSource{forVar, collection} = do
+          collectionTyped <- doTypecheck collection
+          let collectionType = AST.getType collectionTyped
+          unless (isRefType collectionType || isArrayType collectionType || isRangeType collectionType) $
+             pushError collection $ NonIterableError collectionType
+          let forVarType = getInnerType collectionType
+          return (forVar, forVarType)
+
+        getInnerType collectionType
```
kikofernandez, line 1777: you should handle errors. That is, what if this function receives an `IntLiteral`. The compiler would crash.
One way to fix this is to continue with the guards and make sure that the compiler does not crash and throws an error. This will help future compiler writers to know that an `IntLiteral` was not expected. There may be other types that are not expected, not only the `IntLiteral`

kikofernandez (changes requested):
Overall, it looks really good. There are only minor comments on code style. The haskell code seems good.

src/ir/AST/Util.hs
ElieOaks: Was not an issue

critik
```diff
@@ -0,0 +1,9 @@
+The first slide after introduction: Have a picture?
```
kikofernandez: wrong file? This file is not needed here.

forComp/output.enc
```diff
@@ -0,0 +1,15 @@
+typeCheckSource sourceType fors@(ForSource{fsTy, collection}) = do
```
kikofernandez: this is Haskell but uses the Encore file extension.
I am not sure what this is :)

forComp/tests/preliminaryFailSyntaxTests.enc
```diff
@@ -0,0 +1,45 @@
+--These are preliminary for-comprehension tests
```
kikofernandez: Encore does not yet return multiple errors.
Could you instead move each of these tests to their own files (and create their own `.fail` files as well)?

forComp/tryEquals.enc
```diff
@@ -0,0 +1,36 @@
+active class Equal
```
kikofernandez: I am not sure on what is this file testing.
Thanks

forComp/tryEquals.enc
```diff
@@ -0,0 +1,36 @@
...
+    var instance = new Equal()
+    var b = instance!isEqual(n)
+    print("They are equal? {}\n", get(b))
+
+    if ([1, 2, 3].isEqual([1, 2, 3])) then
+      print("It does not work!")
+    end
+
+  end
+end
```
kikofernandez: There is a `forComp/tryFor` binary. Binaries should not be included in git :)

forComp/tryFor.enc
```diff
@@ -0,0 +1,55 @@
...
+    list2.append("katt")
+    list2.append("katt")
+
+    var acc = 0
+    useList(for elem <- list, el <- list2 do
+                    (elem, el)
+            end)
+
+
+    {-var list  = new LinkedList[int]()
```
kikofernandez: is these lines are not tested, please remove them

forComp/tryInt.enc
```diff
@@ -0,0 +1,24 @@
+active class TryInt
```
kikofernandez: are you using this file for anything?

modules/standard/Collections/Mutable/Functor.enc
```diff
@@ -1,10 +1,7 @@
--- JOY for-comprehension
 module Functor
 
-import Collections.Mutable.Collection
-
 local trait Functor[t]
```
kikofernandez, line 4: A functor, as defined in Haskell, is only your `map` operation.
We are not using Haskell concepts, though, so feel free to keep it as is.
This is just a FYI :)

modules/standard/Data/Array.enc
```diff
@@ -85,8 +85,10 @@ end
 -- elements of arr
 fun map[a,b](f : a -> b, arr : [a]) : [b]
   val ret = new [b](|arr|)
-  for i <- [0 .. |arr|-1] do
```
kikofernandez, line 88: can you make this work with for-comprehensions?

modules/standard/Data/Array.enc
```diff
@@ -99,19 +101,20 @@ end
 fun flatMap[a, b](f : a -> [b], arr : [a]) : [b]
   val size = |arr|
   var ret = new [b](0)
-  repeat i <- size do
+  var i = 0
```
kikofernandez, line 102: can you make this work with for-comprehensions?

modules/standard/Data/Array.enc
```diff
@@ -161,7 +164,7 @@ fun contains[t](arr : [t], to_find : t) : bool
   for x <- arr do
     if x == to_find then
       retval = true
-      break
+      -- break TODO: fix break!
```
kikofernandez: is this `break` part of your work and you have plans for fixing this?

modules/standard/Ranges.enc
```diff
@@ -0,0 +1 @@
+module Ranges
```
kikofernandez: I guess this is required as a file for some reason.
Could you remind me of the reason?

modules/standard/String.enc
```diff
@@ -506,3 +507,34 @@ read class String : Id + Hashable + Eq[String](compare(),length,cstring)
     end
   end
 end
+
+-- TODO: put this class in a seperate file
```
kikofernandez: Follow `TODO`?

modules/standard/String.enc
```diff
@@ -506,3 +507,34 @@ read class String : Id + Hashable + Eq[String](compare(),length,cstring)
...
+  def foreach(f : local ((int) -> unit)) : unit
+    var current = this.start
+    while (current <= this.stop) do
+      f(current)
+      current += this.step
+    end
+  end
+end
+
+-- TODO: put this class in a seperate file
```
kikofernandez: Follow `TODO`?

src/back/CodeGen/Expr.hs
```diff
@@ -314,7 +313,7 @@ instance Translatable A.Expr (State Ctx.Context (CCode Lval, CCode Stat)) where
       let exitCall = Call (Nam "exit") [narg]
       return (unit, Seq [Statement targ, Statement exitCall])
 
-  translate abort@(A.Abort {A.args = []}) = do
+  translate abort@(A.Abort {A.args}) = do -- TODO:   translate abort@(A.Abort {A.args = []}) = do
```
kikofernandez: Is this work-in-progress?

src/back/CodeGen/Expr.hs
```diff
@@ -677,6 +676,82 @@ instance Translatable A.Expr (State Ctx.Context (CCode Lval, CCode Stat)) where
...
+        (name, step, src) = getOldFor $ head sources
+
+      indexVar <- Var <$> Ctx.genNamedSym "index"
+      eltVar   <- Var <$> Ctx.genNamedSym (show name)
+      startVar <- Var <$> Ctx.genNamedSym "start"
+      stopVar  <- Var <$> Ctx.genNamedSym "stop"
+      stepVar  <- Var <$> Ctx.genNamedSym "step"
+      srcStepVar <- Var <$> Ctx.genNamedSym "src_step"
+
+      (srcN, srcT) <- if A.isRangeLiteral src
```
kikofernandez: This implementation is not clear to me. Your implementation may assign `undefined` to the `srcN` (source name) variable. This means that in line 701, you could be creating a `Call rangeStart [undefined]`. 
What would be the meaning of this translation? Or from the other angle, what does a developer see if we generate such code?

src/front/TopLevel.hs
```diff
@@ -9,6 +9,8 @@ file I/O.
 
 module Main where
 
+import Debug.Trace
```
kikofernandez: remove when you have finished debugging

src/front/TopLevel.hs
```diff
@@ -338,22 +341,23 @@ main =
...
+
+       verbose options "== Optimizing =="
+       let optimizedTable = fmap optimizeProgram capturecheckedTableTwo
 
        verbose options "== Generating code =="
        let (mainDir, mainName) = dirAndName sourceName
            mainSource = mainDir </> mainName
        let fullAst = setProgramSource mainSource $
-                     compressProgramTable optimizedTable
+                     compressProgramTable typecheckedTableTwo
```
kikofernandez: shouldn't this be the `optimisedTable`?

src/ir/AST/AST.hs
```diff
@@ -19,6 +19,8 @@ import Identifiers
 import Types
 import AST.Meta as Meta hiding(Closure, Async)
 
+import Debug.Trace
```
kikofernandez: removed?

src/opt/Optimizer/Optimizer.hs
```diff
@@ -1,16 +1,14 @@
...
-
 import Identifiers
 import AST.AST
 import AST.Util
 import qualified AST.Meta as Meta
 import Types
 import Control.Applicative (liftA2)
-import Data.Maybe
+
+import Debug.Trace
```
kikofernandez: to be removed?

src/opt/Optimizer/Optimizer.hs
```diff
@@ -179,91 +182,4 @@ forwardGeneral = extend forwardGeneral'
...
-  Assign {emeta = meta,
-          lhs = left,
-          rhs = right}
-
-intoFieldAccess met object nam =
-  FieldAccess{ emeta = met,
-               target = object,
-               name = nam}
--- intoInit --??
+forwardGeneral' e = e
```
kikofernandez: is this indented at the right level?

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+-- for x <- listA, y <- listB, z <- ListC do
+--      fun
+-- end
+--
+-- into listA.flatMap(listB.flatMap(listC.map(fun)))
+forDesugared :: Expr -> Expr
+forDesugared e@For{emeta, sources, body} =
+  let n = length sources
+      collectionType = getType $ collection $ head sources
+      callNameList = if (not (AST.AST.isCaptured e)) || (unitType == getType body) || (isRangeObjectType collectionType)
```
kikofernandez: line a bit too long. try to put parts of it in a new line

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+nestCalls meta (name:restOfNames) (fs:restFS) body elemType =
+  let nestedCall = intoCall meta name fs body elemType
+  in nestCalls meta restOfNames restFS nestedCall elemType
+
+intoCall :: Meta.Meta Expr -> Name -> ForSource -> Expr -> Type -> Expr
+intoCall met callName ForSource{fsName, fsTy, collection} bodyOrMethodCall elemType =
+  if isRefType (getType collection)
+  then let param = [intoParam met Val fsName fsTy]
+           arguments = [intoClosure met param Nothing bodyOrMethodCall]
+           elemT = if callName == Name "foreach" -- this feels iffy
```
kikofernandez: since you have written a bunch of time `Name "foreach"`, maybe it makes sense to define it and create a helper function, `isForeach e`, e.g.
```haskell
foreachName = Name "foreach"
isForeach e = e == foreachName
```

kikofernandez: my feeling is that these kind of constants are written somewhere in the code base

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+                   then []
+                   else [elemType]
+       in  intoMethodCall met elemT collection callName arguments
+   else let param = [intoParam met Val fsName fsTy]
+            arguments = [intoClosure met param Nothing bodyOrMethodCall] ++ [collection]
```
kikofernandez: maybe better:
```haskell
[intoClosure met param Nothing bodyOrMethodCall, collection]
```

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+            elemT =  if callName == Name "foreach" -- this feels iffy
```
kikofernandez: suggestion from two comments above

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+       in   intoFunctionCall met elemT name arguments
+
+boxForInSeq :: Expr -> Expr
+boxForInSeq e@Seq{emeta, eseq} =
+  let newEseq = boxFor eseq []
+  in  e{eseq = newEseq}
+  where
+    boxFor :: [Expr] -> [Expr] -> [Expr]
+    boxFor [] newEseq =  newEseq
+    boxFor (ex:expr) newEseq
```
kikofernandez: I believe this code is the same as:
```
boxFor [] newEseq                   =  newEseq
boxFor ((For{}):expr) newEseq = newEseq ++ [forBoxed ex (boxFor expr [])]
boxFor (ex:expr) newEseq        = boxFor expr (newEseq ++ [ex])
```
Maybe this is more readable than calling that function.

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+      | otherwise = boxFor expr (newEseq ++ [ex])
+    isFor For{} = True
+    isFor _ = False
+
+
+forBoxed :: Expr -> [Expr] -> Expr
+forBoxed for@For{emeta, sources, body} postForExpr =
+  let listOfVar = getVariables body
+      listOfVarNames = map getVarName listOfVar
+      getVarName VarAccess{qname} = qnlocal qname
```
kikofernandez: do you use this function `getVarName VarAccess{qname} = qnlocal qname`?

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+      unBoxed = unBox listOfVar
+      desugaredForWithFieldAccBody = forDesugared $ varBodyToFieldBody for [] listOfVarNames
+      bodyforLetBoxes = intoSeq emeta (desugaredForWithFieldAccBody:unBoxed ++ postForExpr)
+      boxLet = boxVar emeta listOfVar bodyforLetBoxes
+      newSeq = intoSeq emeta [boxLet]
+  in  newSeq
+
+getVariables :: Expr -> [Expr]
```
kikofernandez: my feeling is that if `getVariables` is only used in a function, maybe it makes more sense to create a locally scoped function, using `where`, where it is actually used.

src/opt/Optimizer/TypedDesugarer.hs
```diff
@@ -0,0 +1,257 @@
...
+getVariables body = removeDuplicates (fst (filterVar body)) [] []
+  where
+    removeDuplicates :: [Expr] -> [Name] -> [Expr] -> [Expr]
+    removeDuplicates [] _ finalList = finalList
+    removeDuplicates (e@VarAccess{qname}:expr) listOfNames finalList
+      | (qnlocal qname) `elem` listOfNames = removeDuplicates expr listOfNames finalList
+      | otherwise = removeDuplicates expr ((qnlocal qname):listOfNames) (e:finalList)
+    removeDuplicates (_:expr) listOfNames finalList = undefined
+
+filterVar :: Expr -> ([Expr], [Name])
```
kikofernandez: same for this one, but upon your decision

src/tests/encore/forward/forwardArgInClosure_src/Foo.encore.c
```diff
@@ -0,0 +1,186 @@
+#include "header.h"
```
kikofernandez: this file should not be here

src/tests/encore/forward/forwardArgInClosure_src/Main.encore.c
```diff
@@ -0,0 +1,360 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardCoverage_src/Foo.encore.c
```diff
@@ -0,0 +1,155 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardCoverage_src/Main.encore.c
```diff
@@ -0,0 +1,380 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardFieldAccess_src/Foo.encore.c
```diff
@@ -0,0 +1,107 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardFieldAccess_src/Main.encore.c
```diff
@@ -0,0 +1,363 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardInClosure_src/Main.encore.c
```diff
@@ -0,0 +1,235 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardInnerForward_src/Foo.encore.c
```diff
@@ -0,0 +1,107 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardInnerForward_src/Main.encore.c
```diff
@@ -0,0 +1,362 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardPreexistingFutureFunctionCall_src/Foo.encore.c
```diff
@@ -0,0 +1,170 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardPreexistingFutureFunctionCall_src/Main.encore.c
```diff
@@ -0,0 +1,360 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardPreexistingFuture_src/Foo.encore.c
```diff
@@ -0,0 +1,185 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/forward/forwardPreexistingFuture_src/Main.encore.c
```diff
@@ -0,0 +1,360 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/io/FileTest.tmp
```diff
@@ -0,0 +1,2 @@
+Hello World!
```
kikofernandez: ?

src/tests/encore/linear/burying_src/Main.encore.c
```diff
@@ -0,0 +1,117 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/optaccess/linear_src/Main.encore.c
```diff
@@ -0,0 +1,117 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/optaccess/optAccessMethod_src/Main.encore.c
```diff
@@ -0,0 +1,111 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/optaccess/optAccess_src/Main.encore.c
```diff
@@ -0,0 +1,191 @@
+#include "header.h"
```
kikofernandez: this should not be here

src/tests/encore/par/extract.enc
```diff
@@ -282,6 +282,7 @@ active class Test
       end
       for v <- arr_passive_join do
         if v.tag.eq("pass") then
+          print("{}\n", v)
```
kikofernandez: why?

src/tests/encore/stdlib/Data/BoxedTest.enc
```diff
@@ -1,8 +1,8 @@
...
-import Boxed.Immutable.Bool
-import Boxed.Immutable.Unit
+import Boxed.Integer
+import Boxed.Char
+import Boxed.Real
+import Boxed.Bool
+import Boxed.Unit
 
 active class Main
   def main() : unit
```
kikofernandez: the binary `Count` should not be here

src/types/Typechecker/Typechecker.hs
```diff
@@ -1857,7 +1876,7 @@ instance Checkable Expr where
                   (length expectedTypes) (length args)
       eArgs <- mapM typecheck args
       matchArguments args expectedTypes
-      return $ setType bottomType abort{args=([]::[Expr])}
+      return $ setType bottomType abort{args = eArgs} --args=([]::[Expr])} TODO: is this allowed?
```
kikofernandez: is the comment to address or to remove?

src/types/Typechecker/Typechecker.hs
```diff
@@ -2237,7 +2256,7 @@ matchTypes expected ty
         bindings <- matchArgs (getTypeParameters expected) (getTypeParameters ty)
           `catchError` (\case
                          TCError (TypeMismatchError _ _) _ ->
-                             tcError $ TypeMismatchError ty expected
+                             tcError $ TypeMismatchError $ trace "2259" $ ty expected
```
kikofernandez: remove the trace when you are not debugging anymore

kikofernandez (changes requested):
Overall, the changes look mostly good. There is quite some noise by files that should not be there, binaries that have been added, etc. In general, the comments should be pretty straightforward to address. If you have questions, continue in the local thread :)

Thanks for your work!

modules/standard/Data/Array.enc
ElieOaks: I would rather avoid using for-comprehension inside of map, flatMap and foreach, due to the potential endless compilation loop it could result in. Unless I used it with a range or some other kind of data strucutre. But even then, for performance sake, I would rather use a while, or repeat loop inside of these methods.

